PR #1, opened 2015-12-29
Done

onyoo:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,129 @@ class ApplicationController < Sinatra::Base
...
+  get "/" do
+    redirect "/tweets"
+  end
+
+  get "/tweets" do
+    if logged_in?
+      @fweets = Tweet.all
+      @posts = []
+      @fweets.each do |f|
+        @posts << {User.find(f.user_id).username => f.content, user_id: f.user_id, id: f.id}
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,129 @@ class ApplicationController < Sinatra::Base
...
+      @content = params[:content]
+      Tweet.create(content: @content, user_id: session[:id])
+    else redirect "/login"
+    end
+    redirect "/tweets"
+  end
+
+  get "/users/:slug" do
+      @user = User.find_by(username: params[:slug])
+      @fweets = Tweet.where(user_id: @user.id)
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,129 @@ class ApplicationController < Sinatra::Base
...
+        Tweet.delete(params[:id])
+        redirect "/tweets"
+      end
+    else redirect "/login"
+    end
+  end
+
+  get "/tweets/:id/edit" do
+    @fweet = Tweet.find(params[:id])
+    if logged_in? && @fweet.user_id == session[:id]
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,129 @@ class ApplicationController < Sinatra::Base
...
+      @fweet.save
+      redirect '/tweets'
+    else redirect "/tweets/#{@fweet.user_id}/edit"
+    end
+  end
+
+private
+  helpers do
+    def logged_in?
+      !!session[:id]
```
aviflombaum: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,129 @@ class ApplicationController < Sinatra::Base
...
+    end
+
+    def current_user
+      User.find(session[:id])
```
aviflombaum: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

aviflombaum:
Left you some notes (repetitive at times). Good job!

onyoo:
Thanks! Your point about memoization is great! I will use it in my current project, and in the future.

PR #4, opened 2016-01-02
Done

jakebrady5:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,173 @@ class ApplicationController < Sinatra::Base
...
+      redirect '/tweets'
+    else
+      erb :"users/login"
+    end
+  end
+
+  post '/login' do
+    @user = User.find_by_username(params[:username])
+    session[:id] = @user[:id]
+    if @user.password == params[:password]
```
aviflombaum: You should be using [`has_secure_password`](http://api.rubyonrails.org/classes/ActiveModel/SecurePassword/ClassMethods.html#method-i-has_secure_password) to give user's passwords and be able to authenticate against them.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,173 @@ class ApplicationController < Sinatra::Base
...
+
+  end
+
+  get '/tweets/:id/edit' do
+
+    if session[:id] == nil
+      redirect '/login'
+    end
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:id] == @tweet.user_id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,173 @@ class ApplicationController < Sinatra::Base
...
+  private
+
+
+
+    def is_logged_in?
+      session[:id] != nil
+    end
+
+    def current_user
+      User.find_by_id(session[:id])
```
aviflombaum: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,173 @@ class ApplicationController < Sinatra::Base
...
+    @tweets = Tweet.all
+    erb :"users/show_user"
+  end
+
+  private
+
+
+
+    def is_logged_in?
+      session[:id] != nil
```
aviflombaum: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/models/user.rb
```diff
@@ -0,0 +1,9 @@
+class User < ActiveRecord::Base
+
```
aviflombaum: You should be using [`has_secure_password`](http://api.rubyonrails.org/classes/ActiveModel/SecurePassword/ClassMethods.html#method-i-has_secure_password) to give user's passwords and be able to authenticate against them.

aviflombaum:
Left you some comments, great job @jakebrady5 !

PR #5, opened 2016-01-07
Done

JVThomas:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,157 @@
...
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,157 @@
...
+        redirect to "/tweets"
+      end
+    else
+      redirect to "/login"
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,157 @@
...
+      end
+    else
+      redirect to "/login"
+    end
+  end
+
+  post "/tweets/:id/delete" do
+    if logged_in?
+      @tweet= Tweet.find(params[:id])
+      if @tweet.user_id == session[:id]
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,157 @@
...
+      erb :"tweets/new"
+    else
+      redirect to "/login"
+    end
+  end
+
+  post "/tweets" do
+    if logged_in?
+      if !params["content"].empty?
+        @new_tweet = Tweet.create(content: params["content"], user_id: session[:id])
```
randallreedjr: Prefer using `current_user.id` over `session[:id]` to ensure user is valid and findable.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,157 @@
...
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "upupdowndownleftrightleftrightabselectstart"
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  get '/signup' do
```
randallreedjr: This controller is overloaded. Consider separating your concerns by splitting some of these actions into separate controllers (i.e. a users_controller and a tweets_controller).

randallreedjr:
@JVThomas Nice job! Left some comments for your review. Closing this PR.

PR #6, opened 2016-01-14
Finished project

Lukeghenco:
Me and Philip finished this project as a team.


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+
+
+
+
+  helpers do
+  	def is_logged_in?
+  		!!session[:id]
+  	end
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+
+  	def current_user
+  		User.find(session[:id])
+  	end
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+  get '/login' do
+    if is_logged_in?   
+      redirect to '/tweets'
+    else
+       erb :'users/login'
+    end
+  end
+
+  get '/tweets' do 
+    @tweets = Tweet.all
```
randallreedjr: Wait to set `@tweets` until after validating log in to save an unnecessary call to the database.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+    if is_logged_in?
+      @user = User.find_by(session[:id])
```
randallreedjr: Reuse `current_user` method here.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "  password_security"
   end
 
+  get '/' do 
+    erb :index
+  end
+
+  get '/signup' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+      erb :'tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/new' do 
+    if params["content"] != ""
+      @tweet = Tweet.create(:content => params["content"])
+      @tweet.user = User.find(session[:id])
```
randallreedjr: Utilize `current_user` here as well.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+      @tweet = Tweet.find(params[:id])
+      erb :'tweets/edit_tweet'
+    else 
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/:id' do
+    @tweet = Tweet.find_by(params[:id])
+    if is_logged_in?
```
randallreedjr: You are checking to make sure a user is logged in, but are you verifying that user has access to update this tweet?

randallreedjr:
Nice job @Lukeghenco! Left some comments, closing this PR.

PR #9, opened 2016-01-19
Done

adamgfraser:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,126 @@ class ApplicationController < Sinatra::Base
...
+    redirect to '/login'
+  end
+
+  get '/users/:slug' do
+    @user = user.find_by_slug(params[:slug])
+    erb :'users/show'
+  end
+
+  get '/tweets' do
+    if session[:id]
```
aviflombaum: This should be a call to `logged_in?` no?

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,126 @@ class ApplicationController < Sinatra::Base
...
+      erb :"tweets/show_tweet"
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if session[:id]
+      @tweet = Tweet.find(params[:id])
+      if session[:id] == @tweet.user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,126 @@ class ApplicationController < Sinatra::Base
...
+    if tweet.save
+      redirect to "/tweets/#{params[:id]}"
+    else
+      redirect to "/tweets/#{params[:id]}/edit"
+    end
+  end
+
+  post '/tweets/:id/delete' do
+    tweet = Tweet.find(params[:id])
+    if session[:id] == tweet.user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

aviflombaum:
Great job @adamgfraser left you some notes.

PR #12, opened 2016-01-23
Done

coffeeexistence:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,22 @@
...
+require 'pry'
 
 class ApplicationController < Sinatra::Base
 
   configure do
+    enable :sessions unless test?
+    set :session_secret, "secret"
+  end
+
+  include Helpers
```
aviflombaum: This is not a correct pattern. Sinatra provides ApplicationController level helpers through a [`helpers`](http://www.sinatrarb.com/intro.html#Helpers) macro. You would not need to call the methods defined in the `helpers` block with a module prefix as you do now `Helpers.logged_in?`.

app/controllers/helpers.rb
```diff
@@ -0,0 +1,55 @@
+module Helpers
```
aviflombaum: Beyond being an incorrect pattern, this `Helpers` module is responsible for too much. It knows about logging people in and checking if people can edit tweets. These functionalities should be in different concerns or modules and correctly mixed into `ApplicationHelper` with the `helpers` macro described above.

app/controllers/helpers.rb
```diff
@@ -0,0 +1,55 @@
...
+	def self.available
+		true
+	end
+
```
aviflombaum: The indentation here is way off and should be corrected. Take pride in the aesthetics of your code. 2 spaces, no tabs.

app/controllers/helpers.rb
```diff
@@ -0,0 +1,55 @@
...
+	def self.can_edit?(session, tweet)
+		#binding.pry
+		if session[:id]
+			tweet.user.id==User.find(session[:id]).id
+		else
+			false
+		end
+	end
+
+	def self.modify_tweet(tweet, params)
```
aviflombaum: You've reimplemented `tweet.update({:content => params[:content])` with this helper. Look at ActiveRecord [`update`](http://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-update) and the other ways of [updating objects with activerecord](http://www.davidverhasselt.com/set-attributes-in-activerecord/).

app/controllers/helpers.rb
```diff
@@ -0,0 +1,55 @@
...
+		#puts "ALREADY LOGGED IN" if !!(session[:id])
+		!!(session[:id])
+	end
+
+	def self.signup_params_valid?(p)
+		(p[:username]=="" || p[:password]=="" || p[:email]=="") ? false : true
+	end
+
+	def self.current_user(session)
+		User.find(session[:id])
```
aviflombaum: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/controllers/helpers.rb
```diff
@@ -0,0 +1,55 @@
...
+
+	def self.modify_tweet(tweet, params)
+		#binding.pry
+		tweet.content=params[:content]
+		tweet.save
+
+	end
+
+	def self.valid_login(params)
+		!!User.find_by(username: params[:username], password: params[:password])
```
aviflombaum: It's better to rely on the successful call to `current_user`. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/helpers.rb
```diff
@@ -0,0 +1,55 @@
...
+	end
+
+	def self.login(user, session)
+		#binding.pry
+		#puts "Loggin in user #{user.id}"
+		session[:id]=user.id
+	end
+
+	def self.login_by_params(params, session)
```
aviflombaum: A lot of these methods seem redundant.

I would think you only need `logged_in?` and `current_user` really.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,103 @@
+class TweetsController < ApplicationController
+	get '/tweets' do
```
aviflombaum: Indent correctly.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+			erb :'tweets/view_tweet'
+		else
+			redirect '/login'
+		end
+		
+	end
+
+	post '/tweets/:slug/delete' do  
+		#binding.pry
+		@tweet = Tweet.find(params[:slug].to_i)
```
aviflombaum: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/models/user.rb
```diff
@@ -0,0 +1,10 @@
+class User < ActiveRecord::Base
+	has_many :tweets
```
aviflombaum: You should be using [`has_secure_password`](http://api.rubyonrails.org/classes/ActiveModel/SecurePassword/ClassMethods.html#method-i-has_secure_password) to give user's passwords and be able to authenticate against them.

db/migrate/20160122052547_create_users.rb
```diff
@@ -0,0 +1,9 @@
+class CreateUsers < ActiveRecord::Migration
+  def change
+  	create_table :users do |t|
+  		t.string :username
+  		t.string :email
+  		t.string :password
```
aviflombaum: Should be `password_digest` for the `has_secure_password` functionality and pattern.

aviflombaum:
Good job on this, I left you some comments. I'd try to refactor it using the `helpers` method in Sinatra and `has_secure_password`. Additionally, the code is a bit sloppy with indentation and lots of left in debugging.

PR #16, opened 2016-02-03
Done

ryuichi7:


app/controllers/application_controller.rb
```diff
@@ -3,8 +3,24 @@
...
+  end
+
+  helpers do
+		def logged_in?
+			!!session[:id]
+		end
+
+		def current_user
+			User.find_by_id(session[:id])
+		end
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,24 @@
...
   end
 
+  get '/' do
+  	erb :index
+  end
+
+  helpers do
+		def logged_in?
+			!!session[:id]
+		end
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+			@tweet = Tweet.find_by_id(params[:id])
+			erb :'tweets/show'
+		else
+			redirect 'login'
+		end
+	end
+
+	get '/tweets/:id/edit' do
+		if logged_in?
+			@tweet = Tweet.find_by_id(params[:id])
```
randallreedjr: Shouldn't you check to ensure the user owns the tweet attempting to edit it?

randallreedjr:
@ryuichi7 Nicely done! I left you just a couple comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #18, opened 2016-02-03
Done

mmmeyers:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,164 @@
...
+      redirect '/login'
+    else
+      redirect '/'
+    end
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by(username: params[:username])
+    @tweets = Tweet.all
+    erb :'users/show_user'
```
randallreedjr: Looks like this view is missing

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,164 @@
...
+    @tweet = Tweet.find_by(id: params[:id])
+    if logged_in?
+      erb :'/tweets/show_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    @tweet = Tweet.find_by(id: params[:id])
```
randallreedjr: Shouldn't you check that the logged in user owns the tweet before displaying the edit form?

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,164 @@
...
+#  "email"=>"skittles@aol.com",
+#  "password"=>"rainbows"}
+
+# => #<User:0x007faae80b9038
+#  id: 1,
+#  username: "skittles123",
+#  email: "skittles@aol.com",
+#  password: "rainbows">
 
 class ApplicationController < Sinatra::Base
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,164 @@
...
+      @tweet.content = params[:content]
+      @tweet.save
+      redirect '/tweets'
+    end
+  end
+
+  post '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by(:id => params[:id])
+      @tweet.delete if session[:id] == @tweet.user_id
```
randallreedjr: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `session[:id]`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

randallreedjr:
Great job @mmmeyers! I left you a few comments, feel free to reply. Closing this PR.

PR #19, opened 2016-02-04
Done

ShanaLMoore:


app/models/user.rb
```diff
@@ -0,0 +1,9 @@
+class User < ActiveRecord::Base
```
randallreedjr: You should be using [`has_secure_password`](http://api.rubyonrails.org/classes/ActiveModel/SecurePassword/ClassMethods.html#method-i-has_secure_password) to give users passwords and be able to authenticate against them.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,169 @@ class ApplicationController < Sinatra::Base
...
+      redirect '/tweets'
+    elsif @user.password == params[:password]
+      session[:id] = @user.id
+      redirect '/tweets'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,169 @@ class ApplicationController < Sinatra::Base
...
+  private
+
+
+
+    def is_logged_in?
+      session[:id] != nil
+    end
+
+    def current_user
+      User.find_by_id(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,169 @@ class ApplicationController < Sinatra::Base
...
+    @tweets = Tweet.all
+    erb :"users/show_user"
+  end
+
+  private
+
+
+
+    def is_logged_in?
+      session[:id] != nil
```

```ruby
def logged_in?
  !!current_user
end
```

app/views/users/show_user.erb
```diff
@@ -0,0 +1,11 @@
+<h1>Your Fweets, <%="#{@user.username}"%>!</h1>
+
+<ul>
+  <% @tweets.each do |a| %>
+    <% if a.user_id == @user.id %>
```
randallreedjr: I would move this check into the controller. It would be more efficient to verify the association through ActiveRecord than repeating the validation for each tweet in this loop.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,169 @@ class ApplicationController < Sinatra::Base
...
+
+  end
+
+  get '/tweets/:id/edit' do
+
+    if session[:id] == nil
+      redirect '/login'
+    end
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:id] == @tweet.user_id
```
randallreedjr: Good job checking that the tweet belongs to the user before displaying the edit form.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,169 @@ class ApplicationController < Sinatra::Base
...
+      @tweet.save
+      redirect "/tweets/#{params[:id]}"
+    end
+
+  end
+
+  post '/tweets/:id/delete' do
+
+    @tweet = Tweet.find_by_id(params[:id]) #finds correct tweet
+    if session[:id] == @tweet.user_id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,169 @@ class ApplicationController < Sinatra::Base
...
+    end
+    @user = User.find_by(username: params[:username])
+    if @user == nil
+      @user = User.create(username: params[:username], email: params[:email], password: params[:password])
+      @user.save
+
+      session[:id] = @user.id
+
+      redirect '/tweets'
+    elsif @user.password == params[:password]
```
randallreedjr: I like that you allow an existing user to login via the signup form. This is a convenience that I appreciate when it is present on websites I use.

randallreedjr:
Great job @ShanaLMoore! I left you a few comments, feel free to respond. Closing this PR.

PR #20, opened 2016-02-04
Done

c1505:


app/controllers/application_controller.rb
```diff
@@ -4,7 +4,147 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/new' do
+    if logged_in?
+      erb :"/tweets/create_tweet"
+    else
+      redirect '/login'
+    end
+  end
+  
+  post '/tweets' do
+    if !params[:content].empty?
```
randallreedjr: Do you need to check that user is logged in here?

app/controllers/application_controller.rb
```diff
@@ -4,7 +4,147 @@ class ApplicationController < Sinatra::Base
...
+      @tweet = Tweet.find(params[:id])
+      erb :"/tweets/show_tweet"
+    else
+      redirect '/login'
+    end
+  end
+  
+  get '/tweets/:id/edit' do 
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
```
randallreedjr: Shouldn't a user only be able to edit their own tweets?

app/controllers/application_controller.rb
```diff
@@ -4,7 +4,147 @@ class ApplicationController < Sinatra::Base
...
+    else
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+
+  end
+
+  delete '/tweets/:id/delete' do 
+    tweet = Tweet.find(params[:id])
+    user = current_user
+    if tweet.user_id == user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -4,7 +4,147 @@ class ApplicationController < Sinatra::Base
...
+    redirect '/tweets'
+  end
+      
+  
+  
+
+
+  helpers do 
+    def logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -4,7 +4,147 @@ class ApplicationController < Sinatra::Base
...
+    end
+
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -4,7 +4,147 @@ class ApplicationController < Sinatra::Base
...
+      user = User.new(username: params[:username], email: params[:email], password: params[:password] )
+      user.save
+      session[:user_id] = user.id
+      redirect '/tweets'
+    else
+      redirect '/signup'
+    end
+  end
+
+  get '/tweets' do
```

randallreedjr:
Great job @c1505! I left a few comments, feel free to respond. Closing this PR.

PR #21, opened 2016-02-05
create twitter clone sinatra application

eric-an:


app/controllers/users_controller.rb
```diff
@@ -0,0 +1,52 @@
+class UsersController < ApplicationController
```
randallreedjr: Great job separating functionality into multiple controllers

app/models/concerns/find-by-slug.rb
```diff
@@ -0,0 +1,6 @@
+module FindBySlug
+
+  def find_by_slug(slug)
```
randallreedjr: I think this method could have been added directly to the Slug module

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+    erb :"index"
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+    set :session_secret, "password_security"
   end
 
+  get '/' do
+    erb :"index"
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+      erb :"tweets/new"
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    @user = current_user
+    @tweet = Tweet.create(content: params[:content])
+    @tweet.user = @user
```
randallreedjr: You could do this in the same step.
```
@tweet = Tweet.create(content: params[:content], user: @user)
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+    if logged_in? && @tweet.user == current_user
+      erb :"tweets/edit"
+    else
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id/edit' do
+    @tweet = Tweet.find_by(id: params[:id])
+    if @tweet.update(content: params[:content])
```
randallreedjr: Might be good to double check tweet being updated belongs to current user, as a security issue.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+      redirect to "/tweets/#{@tweet.id}"
+    else
+      redirect to "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by(id: params[:id])
+    if logged_in? && @tweet.user == current_user
```
randallreedjr: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user` association of the tweet against the `current_user`.

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

randallreedjr:
Great job @eric-an! I left you a few comments, feel free to respond. Closing this PR.

PR #22, opened 2016-02-05
Done

benser1:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
+require 'pry'
+
 require './config/environment'
 
+
 class ApplicationController < Sinatra::Base
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
...
+    user = User.find_by(:username => params[:username])
+    if user && user.authenticate(params[:password])
+      session[:id] = user.id
+      redirect '/tweets'
+    else
+      redirect '/login' 
+    end
+  end 
+
+  get '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
...
 
-end
\ No newline at end of file
+  post '/tweets/new' do   
+    if params[:content] == "" 
+      redirect '/tweets/new'
+    else
+      @tweet = Tweet.new(content: params[:content])  
+      @user = current_user
+      @tweet.user_id = @user.id
```
randallreedjr: The tweet creation and user association can both be accomplished in the same step.

``` ruby
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
...
+    @tweet = Tweet.find_by(id: params[:id])
+    if logged_in?
+      erb :'/tweets/show_tweet' 
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do 
+    @tweet = Tweet.find_by(id: params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
...
+       redirect to "/tweets/#{@tweet.id}/edit"
+     else
+       redirect to "/tweets/#{@tweet.id}/edit"
+     end
+   end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+       @tweet = Tweet.find_by_id(params[:id])
+       if session[:id] == @tweet.user_id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
...
+ end
+
+  get '/logout' do 
+    session.clear
+    redirect '/login'
+  end
+
+  helpers do 
+    def current_user
+     User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
...
+    end
+ 
+    def logged_in?
+     !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

randallreedjr:
Great job @benser1! I left a few comments, feel free to respond. Closing this PR.

PR #24, opened 2016-02-08
Done

azheng249:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,133 @@
...
+      redirect "/login"
+    end
+  end
+
+  get '/logout' do
+    session.clear if Helpers.is_logged_in?(session)
+    redirect "/login" if !Helpers.is_logged_in?(session)
+  end
+
+  get '/users' do
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,11 @@
+class Helpers
+
+  def self.current_user(session)
+    User.find(session[:user_id])
```
randallreedjr: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,11 @@
...
+  end
+
+  def self.is_logged_in?(session)
+    !!session[:user_id]
```
randallreedjr: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/models/user.rb
```diff
@@ -0,0 +1,16 @@
...
+  validates :email, presence: true
+  validates :password, presence: true
+  has_secure_password
+
+  def slug
+    self.username.gsub(" ", "-")
+  end
+
+  def self.find_by_slug(slug)
+    self.all.find {|user| user.slug == slug}
```

```ruby
def self.find_by_slug(slug)
  User.find_by(slug: slug)
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,133 @@
...
+      end
+    else
+      redirect "/login" 
+    end  
+  end
+
+  post '/tweets/:id/delete' do
+    if Helpers.is_logged_in?(session)
+      @tweet = Tweet.find(params[:id])
+      if @tweet.user_id == Helpers.current_user(session).id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/views/layout.erb
```diff
@@ -0,0 +1,28 @@
+<html>
+  <head>
+    <title>Fwitter - Flatiron Twitter</title>
+
+    <!-- Stylesheets -->
+    <link href="<%= url('/css/bootstrap.min.css') %>" rel="stylesheet" type="text/css" />
+    <link href="<%= url('/css/style.css') %>" rel="stylesheet" type="text/css" />
```
randallreedjr: Definitely going above and beyond. Awesome!

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,133 @@
...
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
+      redirect "/signup", locals: {message: "Please submit enter all the information to complete signup."}
+    else
+      @user = User.create(
+        username: params[:username],
+        email: params[:email],
+        password: params[:password]
+      )
+      session[:user_id] = @user.id
+      redirect "/tweets", locals: {message: "Welcome to Fwitter. Thank you for signing up."}
```
randallreedjr: I got an error when trying to signup. Is it still working for you?

randallreedjr:
Great job @azheng249! Unfortunately, I was unable to sign up when I tried running your code. I left you a few comments, feel free to respond. Closing this PR.

app/controllers/application_controller.rb
azheng249: Wait what error did you get? I cloned my repo and was able to make a new
account.

randallreedjr: 127.0.0.1 - - [28/Feb/2016:22:34:33 -0500] "POST /signup HTTP/1.1" 303 - 0.0747
D, [2016-02-28T22:34:33.306229 #26586] DEBUG -- :    (0.1ms)  begin transaction
D, [2016-02-28T22:34:33.307061 #26586] DEBUG -- :   SQL (0.2ms)  INSERT INTO "users" ("username", "email", "password_digest") VALUES (?, ?, ?)  [["username", "rreed"], ["email", "rreed@test.com"], ["password_digest", "$2a$10$gqjg1wfIPudGIsLcv05DXekR0PLLTcyOP3a.J0ipcM1Nkl4/Z6/nu"]]
D, [2016-02-28T22:34:33.309637 #26586] DEBUG -- :    (2.4ms)  commit transaction
Unexpected error while processing request: Body yielded non-string value [:locals, {:message=>"Welcome to Fwitter. Thank you for signing up."}]

azheng249: Okay I see the issue. I should change it from redirect to erb :"tweets/index". Thanks for telling me. That's weird, it just redirected me to tweets when I did it.

PR #25, opened 2016-02-08
Done

coltonstaab1:


randallreedjr:
Hi @coltonstaab1. Unfortunately, there were some extra files included in your submission that make it impossible to review your code. Could you please remove the `.bundle` directory and resubmit?

coltonstaab1:
Hi Randall,

Sorry about the delay--just updated my project, so it should be good to go.

Thanks,
Colton

randallreedjr:
Hi @coltonstaab1, it looks like the project still has the .bundle directory.

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,126 @@
...
+	end
 
-end
\ No newline at end of file
+	get '/' do
+		erb :index
+	end
+
+	get '/signup' do
+		if !session[:id].nil?
```
randallreedjr: I don't see a `current_user` or `logged_in?` helper method, but they can be helpful when dealing with authenticated users. A bit more semantic than looking up session variables.

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,126 @@
...
+			redirect '/tweets'
+		end
+		erb :'/users/create_user'
+	end
+
+	post '/signup' do
+		@user = User.create(params)
+		if params[:username] == "" || params[:password] == "" || params[:email] == ""
```
randallreedjr: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :username`) to perform these checks more easily.

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,126 @@
...
+			redirect '/signup'
+		end
+
+		if @user.save
```
randallreedjr: `User.create` should have persisted the user for you. If you want to include an `if @user.save` check, `User.new` might make more sense.

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,126 @@
...
+		end
+	end
+
+	post "/login" do
+		@user = User.find_by(username: params[:username])
+		if @user && @user.authenticate(params[:password])
+			session[:id] = @user.id
+			redirect "/tweets"
+		else
+			redirect "/failure"
```
randallreedjr: Where does this go?

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,126 @@
...
+		end		
+		erb :'/tweets/new'
+	end
+
+	post '/tweets/new' do
+		if params[:content] == ""
+			redirect :'/tweets/new'	
+		end
+		@tweet = Tweet.create(params)
+		@tweet.user_id = session[:id]
```

``` ruby
```

or alternatively

```
```

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,126 @@
...
+		@tweet = Tweet.find_by_id(params[:tweet_id])
+		erb :'/tweets/display_single_tweet'
+	end
+
+	get '/tweets/:tweet_id/edit' do
+		@tweet = Tweet.find_by_id(params[:tweet_id])
+		if session[:id].nil?
+			redirect '/login'
+		end
+		if session[:id] != @tweet.user_id
```
randallreedjr: I rarely see this check, well done!

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,126 @@
...
+		if params[:content] != ""
+			@tweet.content = params[:content]
+			@tweet.save
+		end
+		#redirect '/tweets/:tweet_id'
+	end
+
+	post '/tweets/:tweet_id/delete' do
+		@tweet = Tweet.find_by_id(params[:tweet_id])
+		if session[:id] != @tweet.user_id
```

```ruby
tweet = current_user.tweets.find_by(:id => params[:tweet_id])
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

Note the usage of [`find_by`](http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find-by) vs [`find`](http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find). Had we used `find` here, ActiveRecord would raise an error in the event that we could not find a tweet associated with this user. `find_by`, like `find_by_id`, will allow this action to fail gracefully, simply returning nil if the user is trying to delete a tweet that they do not have access to.

app/models/user.rb
```diff
@@ -0,0 +1,13 @@
+class User < ActiveRecord::Base
+	has_secure_password
+	has_many :tweets
+
+	def slug
+		self.username.downcase.gsub(" ","-")
```
randallreedjr: This is interesting...wouldn't it be possible to have multiple users with the same slug? For example '@john smith' and '@john-smith'.

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,126 @@
...
+		@user = User.find_by(username: params[:username])
+		if @user && @user.authenticate(params[:password])
+			session[:id] = @user.id
+			redirect "/tweets"
+		else
+			redirect "/failure"
+		end
+	end
+
+	get '/tweets' do
```

randallreedjr:
@coltonstaab1 Sorry for the delay in reviewing this submission, glad we got the bundle files straightened out. You did a great job! I left a few notes, feel free to respond with any questions or comments. Closing this PR for now.

PR #26, opened 2016-02-09
Project submitted

Rawradin:


randallreedjr:
@Rawradin Is this submission complete? It seems to be missing some functionality.

Rawradin:
I thought it was complete. I'll take a look at it when I get out of work.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,84 @@
...
+    erb :tweets
+  end
+
+  get '/logout' do 
+    session.clear
+    redirect '/'
+  end
+
+  helpers do 
+    def valid_new_user(params)
```
randallreedjr: I think you could probably rely on the ActiveRecord model validations instead of buidling a custom validator, e.g. `validates :username, presence: true, uniqueness: true`.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,84 @@
...
+    if !logged_in 
+      erb :signup
+    else
+     redirect '/tweets'
+    end
+  end
+
+  post '/signup' do
+    if valid_new_user(params)
+      @user = User.new(username: params[:username],email: params[:email],password_digest: BCrypt::Password.create(params[:password]))
```
randallreedjr: I think if you pass `params[:password]` to the `password` attribute, it should automatically convert it to the password digest.

randallreedjr:
Hi @Rawradin. I don't see how a user can view/create tweets?

PR #27, opened 2016-02-11
Project submitted

caporta:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+  end
+
+  get '/logout' do
+    session.clear if logged_in?
+    redirect '/login'
+  end
+
+  helpers do
+    def current_user
+      @user = User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+    def find_user
+      @user = User.find_by(username: params[:username])
+    end
+
+    def authenticate_user
+      @user.authenticate(params[:password])
+    end
+
+    def user_exists?
+      !!@user
```
randallreedjr: If `current_user` is memoized (see above), this method becomes extraneous.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+    end
+
+    def logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+    session[:user_id] = @user.id
+    redirect '/tweets'
+  end
+
+  get '/logout' do
+    session.clear if logged_in?
+    redirect '/login'
+  end
+
+  helpers do
```
randallreedjr: Very nice suite of helper methods.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
+class TweetsController < ApplicationController
```
randallreedjr: Great job splitting this functionality out to a separate controller.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    if logged_in?
+      @tweets = Tweet.all
+      erb :'tweets/index'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    redirect '/tweets/new' unless user_credentialed?
```
randallreedjr: Is `user_credentialed?` an accurate name for the method if it is being used in this way?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      erb :'tweets/edit'
+    else
+      redirect '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    find_tweet
+    redirect "/tweets/#{@tweet.id}/edit" if params[:content].blank?
+    @tweet.update(content: params[:content])
```
randallreedjr: Shouldn't you verify a tweet belongs to the `current_user` before allowing them to edit?

app/views/tweets/new.erb
```diff
@@ -0,0 +1,13 @@
+<h1>New Tweet:</h1>
+
+<form action="/tweets" method="POST">
+  <input type="hidden" name="user_id" id="hidden" value="<%=current_user.id%>" />
```
randallreedjr: I would recommend against doing this. Even though the field is hidden, it can be manipulated through the browser, allowing a malicious user to create a tweet under a different user's account. Instead, associate the new tweet with the `current_user` in the controller, relying on the session instead of the form field.

app/models/concerns/slugifiable.rb
```diff
@@ -0,0 +1,13 @@
+module Slugifiable
+  module InstanceMethods
+    def slug
+      username.downcase.gsub(" ","-")
+    end
+  end
+
+  module ClassMethods
+    def find_by_slug(slug)
+      all.find { |instance| instance.slug == slug }
```
randallreedjr: Instead of redefining your own `find_by_slug` method, you can utilize the existing ActiveRecord `find_by` method.

```ruby
def find_by_slug(slug)
  find_by(slug: slug)
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    redirect "/tweets/#{@tweet.id}"
+  end
+
+  delete '/tweets/:id' do
+    find_tweet
+    @tweet.delete if tweet_belongs_to_user
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/views/users/show.erb
```diff
@@ -0,0 +1,2 @@
+<h1>User is tweeting!</h1>
+<h2>"tweet tweet tweet"</h2>
```
randallreedjr: Not sure what the point of this view is...lol

randallreedjr:
Great job @caporta! I left a few comments, feel free to respond. Closing this PR.

PR #31, opened 2016-02-13
Done

simque:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,148 @@
...
+      user = User.new(username: params[:username])
+      user.email = params[:email]
+      user.password = params[:password]
+      user.save
+      session[:id] = user.id
+      redirect '/tweets'
+    end
+  end
+
+  get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,148 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    if !params[:content].empty?
+      @tweet = Tweet.find(params[:id])
+      @tweet.content = params[:content]
+      @tweet.save
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,148 @@
...
+      redirect "/tweets/#{@tweet.id}"
+    else
+      @tweet = Tweet.find(params[:id])
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
+    if @tweet.user_id == current_user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,148 @@
...
+    end
+  end
+
+  helpers do
+    def is_logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,148 @@
...
+      session.clear
+      redirect '/login'
+    else
+      redirect '/'
+    end
+  end
+
+  helpers do
+    def is_logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/models/user.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def self.find_by_slug(slug)
+    found = nil
+    User.all.each do |user|
+      if user.slug == slug
+        found = user
+      end
+    end
+    found
```

```ruby
def self.find_by_slug(slug)
  User.find_by(slug: slug)
end
```

app/views/tweets/create_tweet.erb
```diff
@@ -0,0 +1,6 @@
+<h2>Create a New Tweet</h2>
+<form method="POST" action="/tweets">
+  <label>Content:</label>
+  <textarea maxlength="140" name="content"></textarea>
```
randallreedjr: Nice length enforcement

app/views/tweets/edit_tweet.erb
```diff
@@ -0,0 +1,11 @@
+<% if @tweet.user_id != current_user.id %>
+<%= @tweet.content %><br>
+<% end %>
+
+<% if @tweet.user_id == current_user.id %>
```
randallreedjr: Ah I see. This clarifies why you didn't perform the user check in the controller. However, I would still include it there, in case a malicious user attempts to bypass your form.
Also, rather than just display the content for a user who doesn't own the tweet, I would redirect them to the show_tweet view (so that the URL remains accurate).

randallreedjr:
Create job @simque! I left you a few comments, feel free to respond. Closing this PR.

PR #32, opened 2016-02-14
Project submitted

HarlemSquirrel:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
 
-end
\ No newline at end of file
+  get '/' do
+    @session = session
+    @tweets = Tweet.all
+    erb :index
+  end
+
+  ################### users ###################
```
randallreedjr: These comments are nice, but...why not just put the actions in a separate controller (i.e. `users_controller` and `tweets_controller`)?

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+  get '/logout' do
+    session.clear
+    redirect '/login'
+  end
+
+
+  ################ tweets ################
+  get '/tweets' do
+    redirect to '/login' unless Helpers.is_logged_in?(session)
+    @user = User.find(session[:id])
```
randallreedjr: You should be using your `current_user

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+      user.tweets << Tweet.all.last
+      redirect '/tweets'
+    else
+      redirect '/tweets/new'
+    end
+  end
+
+  get '/tweets/:id' do
+    redirect to '/login' unless Helpers.is_logged_in?(session)
+    @session = session
```
randallreedjr: You shouldn't need to store the session in an instance variable.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+    @user = User.find(session[:id])
+    @tweet = Tweet.find_by(id: params[:id])
+    erb :'tweets/edit_tweet'
+  end
+
+  post '/tweets/:id' do
+    tweet = Tweet.find_by(id: params[:id])
+    tweet.update(content: params[:content])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+      redirect "/tweets/#{tweet.id}/edit"
+    else
+      redirect "/tweets/#{tweet.id}"
+    end
+  end
+
+  post '/tweets/:id/delete' do
+    redirect to '/login' unless Helpers.is_logged_in?(session)
+    begin
+      tweet = Helpers.current_user(session).tweets.find(params[:id])
```
randallreedjr: Good use of ActiveRecord associations

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+
+
+end
+
+
+class Helpers
+  def self.current_user(session)
+    User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+    @user = User.find_by_slug(params[:slug])
+    erb :'tweets/tweets'
+  end
+
+
+
+end
+
+
+class Helpers
```
randallreedjr: Stick to one class definition per file.

app/models/user.rb
```diff
@@ -0,0 +1,34 @@
+class User < ActiveRecord::Base
+  has_many :tweets
+  has_many :active_relationships, class_name: "Relationship",
+                                  foreign_key: "follower_id",
+                                  dependent: :destroy
+  has_many :following, through: :active_relationships, source: :followed
+  has_many :passive_relationships, class_name: "Relationship",
+                                   foreign_key: "followed_id",
+                                   dependent: :destroy
+  has_many :followers, through: :passive_relationships, source: :follower
```
randallreedjr: Woh! Very cool!

app/models/user.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  has_secure_password
+
+  def slug
+    self.username.downcase.gsub(" ", "-")
+  end
+
+  def self.find_by_slug(slug)
+      self.all.detect { |instance| instance.username.downcase == slug.gsub("-", " ")}
```

```ruby
def self.find_by_slug(slug)
  User.find_by(slug: slug)
end
```

randallreedjr:
Great job @HarlemSquirrel! I left a few comments, feel free to respond. Closing this PR.

app/controllers/application_controller.rb
HarlemSquirrel: This makes a lot of sense! Thank you!

app/models/user.rb
HarlemSquirrel: Wow, I didn't know about that! Cool!

HarlemSquirrel:
Thank you @randallreedjr for your feedback!

PR #33, opened 2016-02-16
complete.

AdamKing0126:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,28 @@
...
 
-end
\ No newline at end of file
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,28 @@
...
+
+  get '/' do
+    erb :index
   end
 
-end
\ No newline at end of file
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    if @tweet.save
+      redirect '/tweets'
+    else
+      redirect 'tweets/new'
+    end
+  end
+
+  delete '/tweets/:id' do
+    tweet = Tweet.find(params[:id].to_i)
+    redirect '/tweets' if tweet.user_id != current_user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    tweet.delete
+    redirect '/tweets'
+  end
+
+  get '/tweets/:id/edit' do
+    redirect '/login' unless logged_in?
+    @tweet = Tweet.find(params[:id].to_i)
```
randallreedjr: Should you verify the user owns the tweet they are trying to edit, not just that any user is logged in?

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      log_in_user
+      redirect '/tweets'
+    else
+      redirect '/signup'
+    end
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    @tweets = Tweet.all { |tweet| tweet.user_id == @user.id }
```
randallreedjr: Here again, you should be able to use `@user.tweets` instead of checking the foreign key. Also, it is preferable here to use ActiveRecord's `where` method to filter results on the database side, rather than processing the results in Ruby.

randallreedjr:
@AdamKing0126 Great job! I left you a few comments, feel free to respond. Closing this PR for now.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,57 @@
+class TweetsController < ApplicationController
```
randallreedjr: Great job dividing the actions into different controllers.

app/models/user.rb
```diff
@@ -0,0 +1,15 @@
...
+  validates :username, presence: true
+  validates :email, presence: true
+  validates :password, presence: true
+
+  def slug
+    self.username.gsub(/[[:punct:]]/, '').gsub(/\s/, '-')
+  end
+
+  def self.find_by_slug(slug)
+    User.all.find { |user| user.slug == slug }
```
randallreedjr: Unlike in the get '/users/:slug' action in the UsersController, here User.all.find is the appropriate method, since the `slug` attribute doesn't exist in the database.

PR #35, opened 2016-02-17
Done

jsgoldb:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
 
 class ApplicationController < Sinatra::Base
+  
+  register Sinatra::ActiveRecordExtension
 
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "its_a_secret"
```
randallreedjr: http://secrettoeverybody.com/

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+  get '/signup' do
+    if !is_logged_in
+      erb :'registrations/signup'
+    else
+      redirect '/tweets'
+    end 
+  end
+
+  post '/signup' do
+    if params[:username] != '' && params[:password] != '' && params[:email] != ''
```
randallreedjr: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) to perform these checks more easily.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+    erb :'users/index'
+  end
+
+  helpers do
+    def is_logged_in
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    erb :'users/index'
+  end
+
+  helpers do
+    def is_logged_in
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+  get '/tweets/new' do
+    if is_logged_in
+      erb :'tweets/create'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    if !params[:content].empty?
```
randallreedjr:
```
class Tweet < ActiveRecord::Base
  validates_presence_of :content
...
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+      @tweet = Tweet.find(params[:id])
+      erb :'tweets/show'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if is_logged_in
+      @tweet = Tweet.find(params[:id])
```
randallreedjr: Shouldn't we confirm the user owns the tweet before allowing them to edit it?

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+      @tweet.save
+      redirect "/tweets/#{@tweet.id}"
+    else
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  delete '/:id/delete' do
+    @tweet = Tweet.find(params[:id]) 
+    if is_logged_in && @tweet.user_id == current_user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+      session.clear
+      redirect '/login'
+    else
+      redirect '/'
+    end
+  end
+
+
+
+
```

app/models/concerns/slugifiable.rb
```diff
@@ -0,0 +1,19 @@
+module Slugifiable
```
randallreedjr: This is a really nice touch.

capybara-201602171056288179438470.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
```
randallreedjr: What are these capybara files?

randallreedjr:
@jsgoldb Great job! I left a few comments, feel free to respond with any questions or comments. Closing this PR for now.

PR #37, opened 2016-02-21
Project submitted

macgeek30:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
  @current_user ||= User.find(session[:user_id) if session[:user_id]
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+      session.destroy
+      redirect to '/login'
+    else
+      redirect to '/'
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
```
randallreedjr: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:user_id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+    enable :sessions
+    set :session_secret, "fwitter_secret"
   end
 
+  get '/' do
+    #welcome to fwitter!!!
+    erb :index
+  end
+
+  get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+    #index of all tweets
+    if session[:user_id]
```
randallreedjr: You can use your `logged_in?` helper method here.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+      erb :'tweets/create'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    #creates new tweet
+    if params[:content] == ""
+      redirect to "/tweets/new"
```
randallreedjr: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :content`) to perform these types of checks more easily.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+    else
+      user = User.find_by_id(session[:user_id])
```
randallreedjr: aka...`current_user`?

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+      @tweet = Tweet.create(:content => params[:content], :user_id => user.id)
```
```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+      erb :'tweets/show'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do  #load edit form
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
randallreedjr: Good job checking that user has access to update tweet before allowing them to edit it. However, we can make a slight improvement on this by finding tweets that are already only associated to the `current_user`, rather than manually checking the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
@tweet = current_user.tweets.find_by(:id => params[:id])
if @tweet
  erb :'tweets/edit'
else
  redirect "/tweets"
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/:id' do #edit action
+    if params[:content] == ""
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  post '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
+      binding
```
randallreedjr: oops?

db/schema.rb
```diff
@@ -0,0 +1,31 @@
...
+  create_table "tweets", force: :cascade do |t|
+    t.string   "content"
+    t.integer  "user_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+  create_table "users", force: :cascade do |t|
+    t.string   "username"
+    t.string   "email"
```
randallreedjr: email is a nice addition!

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+      redirect to '/tweets'
+    end
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
+      redirect to '/signup'
+    else
+      @user = User.new(:username => params[:username], :email => params[:email], :password => params[:password])
+      @user.save
```
randallreedjr: Probably want to verify user creation succeeded before storing ID in the session (and essentially logging them in).

app/models/user.rb
```diff
@@ -0,0 +1,13 @@
+class User < ActiveRecord::Base
+  has_many  :tweets
+  has_secure_password
+
+  def slug
+    username.downcase.gsub(" ","-")
```
randallreedjr: This is interesting...wouldn't it be possible to have multiple users with the same slug? For example '@john smith' and '@john-smith'.

randallreedjr:
@macgeek30 Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

PR #38, opened 2016-02-22
completed.

matthewpaulmcgowan:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+  get '/signup' do 
+    if logged_in?
+      redirect "/tweets"
+    else
+      erb :"users/create_user"
+    end
+  end
+  
+  post '/signup' do 
+    if params["username"] == "" || params["email"] == ""
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+      redirect "/signup"
+    end
+    
+    @user = User.new(username: params["username"], email: params["email"], password: params["password"])
```
randallreedjr: You can shorthand this with `User.new(params)` and ActiveRecord will extrapolate the fields from the params hash.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+    if @user.save
+      session[:id] = @user.id
+      redirect "/tweets"
+    else
+      redirect "/signup"
+    end
+  end
+  
+  get '/login' do
+    if session[:id].nil?
```
randallreedjr: You can use your `logged_in?` helper method here.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+      erb :"users/login"
+    else
+      redirect "/tweets"
+    end
+  end
+  
+  post '/login' do
+    @user = User.find_by(username: params["username"]) #cannot use current_user here because on the rspec for this, it does not register the session id, which the method uses rspec lines:81-92
```
randallreedjr: I'm not sure I follow. If you are attempting to login, there isn't a current user yet...

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+  end
+  
+  post "/tweets" do
+    if params["content"] == "" 
+      redirect "/tweets/new"
+    end
+    @user = current_user
+    @tweet = Tweet.create(content: params["content"])
+    @tweet.user_id = @user.id 
+    @tweet.save
```

``` ruby
```

or alternatively

```
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+    erb :"users/user_homepage"
+  end
+  
+  get '/tweets/:id' do 
+    @tweet = current_tweet
```
randallreedjr: current_tweet is interesting. I have not seen that before!

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+    end
+  end
+  
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+        redirect "/tweets"
+      end
+    else
+      redirect '/login'
+    end
+  end
+  
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+    end
+
+    def current_user
+      User.find(session[:id])
+    end
+    
+    def current_tweet
+      Tweet.find(params[:id])
```
randallreedjr: Slight preference for [`find_by`](http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find-by) vs [`find`](http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find). When using `find, ActiveRecord will raise an error in the event that it cannot find a tweet for the given id. `find_by` will allow this action to fail gracefully, simply returning nil if the user is trying to access a tweet that does not exist (which could be handled elsewhere in the app).

app/views/tweets/show_tweet.erb
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+  <head>
+  </head>
+  <body>
+    <h2><%=@tweet.content%></h2>
+    <h2><a href="/tweets/<%=@tweet.id%>/edit">Edit Tweet</a></h2>
+    <h2><a href="/users/<%=User.find(@tweet.user_id).slug%>">User Home Page</a></h2>
```
randallreedjr: This `User.find` lookup should really be handled in the controller.

randallreedjr:
@matthewpaulmcgowan Great job! There is a lot of polish here, and some nice usability features. I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+  get '/logout' do
+    if logged_in?
+      session.clear
+      redirect "/login"
+    else
+      redirect '/'
+    end
+  end
+    
+  get '/tweets' do
```

PR #39, opened 2016-02-23
Done

alexbarron:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+      redirect "/login", locals: {message: "Can't log out, you're not logged in."}
+    end
+  end
+
+  get "/users/:slug" do
+    @user = User.find_by_slug(params[:slug])
+    erb :'/users/show'
+  end
+
+  get "/tweets" do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+      redirect "/tweets"
+    else
+      redirect "/login", locals: {message: "Please log in to see that."}
+    end
+  end
+
+
+  helpers do
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+    end
+
+    def logged_in?
+      session[:user_id] ? true : false
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+    elsif params[:content] == ""
+      redirect "/tweets/#{@tweet.id}/edit", locals: {message: "Tweets can't be empty."}
+    else
+      redirect "/login", locals: {message: "Please log in to see that."}
+    end
+  end
+
+  delete "/tweets/:id/delete" do
+    @tweet = Tweet.find(params[:id])
+    if logged_in? && @tweet.user == current_user
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/models/user.rb
```diff
@@ -0,0 +1,14 @@
+class User < ActiveRecord::Base
+  has_many :tweets
+  has_secure_password
+  validates_presence_of :username, :email
```
randallreedjr: Great! First person I have seen use the ActiveRecord validators.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      erb :'/tweets/new'
+    else
+      redirect "/login", locals: {message: "Please log in to see that."}
+    end
+  end
+
+  post "/tweets" do
+    if logged_in? && params[:content] != ""
+      Tweet.create(content: params[:content], user_id: session[:user_id])
```
```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+      erb :'/tweets/show'
+    else
+      redirect "/login", locals: {message: "Please log in to see that."}
+    end
+  end
+
+  get "/tweets/:id/edit" do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
+      erb :'/tweets/edit'
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "password_security"
   end
 
+  get '/' do
+    @user = User.find(session[:user_id]) if session[:user_id]
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+      erb :'/tweets/edit'
+    else
+      redirect "/login", locals: {message: "Please log in to see that."}
+    end
+  end
+
+  post "/tweets/:id" do
+    @tweet = Tweet.find(params[:id])
+    if logged_in? && params[:content] != ""
+      @tweet.update(content: params[:content])
```

app/models/user.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def slug
+    self.username.downcase.gsub(" ", "-")
```
randallreedjr: This is interesting...wouldn't it be possible to have multiple users with the same username and/or slug? For example '@john smith' and '@john-smith'.

randallreedjr:
@alexbarron Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

alexbarron:
Hi @randallreedjr, sorry for the late response! This is a really helpful feedback and I'm incorporating it to re-factor my Sinatra assessment project.

randallreedjr:
Great, glad you found it helpful!

PR #40, opened 2016-02-24
Done

Louiefigz:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,176 @@ class ApplicationController < Sinatra::Base
...
+        session[:user_id] = @user.id
+      
+        redirect '/tweets'
+      else 
+        redirect '/signup'
+      
+    end 
+  end 
+
+  get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,176 @@ class ApplicationController < Sinatra::Base
...
+    redirect '/tweets'
+  end 
+  end 
+
+
+  
+  helpers do 
+
+    def current_user
+      User.find(session["user_id"])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,176 @@ class ApplicationController < Sinatra::Base
...
+    end 
+
+    def is_logged_in?
+       !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,176 @@ class ApplicationController < Sinatra::Base
...
+    redirect '/tweets'
+  else 
+    redirect '/tweets'
+  end 
+  end 
+
+  post '/tweet/:id/delete' do 
+  
+    if is_logged_in? && User.find(Tweet.find(params[:id]).user_id) == User.find(session[:user_id])
+      @tweet = Tweet.find(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,176 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/tweets/new' do 
+      if is_logged_in?
+        erb :'/tweets/create_tweet'
+      else 
+        redirect "/login"
+    end
+  end 
+
+  post '/tweets/new' do
```
randallreedjr: Ideally, this route would be `post '/tweets'`.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,176 @@ class ApplicationController < Sinatra::Base
...
+  end 
+
+  post '/tweets/new' do 
+   if !params[:tweet].present?
+    redirect '/tweets/new'
+  end 
+
+    if is_logged_in? && params[:tweet].present?
+      @user = User.find(session[:user_id])
+      @tweet = Tweet.create(content: params[:tweet], user_id: @user.id)
```
```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,176 @@ class ApplicationController < Sinatra::Base
...
+      if is_logged_in?
+          redirect '/tweets'
+      else 
+      erb :'users/create_user'
+    end 
+  end
+
+  post '/signup' do 
+   
+      if params[:username].present? && params[:email].present? && params[:password].present?
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,176 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/tweets' do 
+ 
+    if is_logged_in?
+        @fweets = Tweet.all
+        @posts = []
+        @fweets.each do |f|
+        @posts << {username: User.find(f.user_id).username, content: f.content, user_id: f.user_id, id: f.id}
+          end
+      @user = User.find(session[:user_id])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,176 @@ class ApplicationController < Sinatra::Base
...
+      redirect '/login'
+    end 
+  end 
+
+  post '/tweets/:id/edit' do 
+  
+    
+  if params[:tweet].present?
+    @tweet= Tweet.find(params[:id])
+    @tweet.update(content: params[:tweet])
```

app/models/concerns/slugifiable.rb
```diff
@@ -0,0 +1,14 @@
+module Sluggable
+  module InstanceMethods
+    def slug
+   
+      username.downcase.gsub(" ", "-")
```
randallreedjr: This is interesting...wouldn't it be possible to have multiple users with the same username and/or slug? For example '@john smith' and '@john-smith'.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,5 @@
+class TweetController < ApplicationController
+
+
+
+end
```
randallreedjr: Your head was in the right place.

randallreedjr:
@Louiefigz Great job! Very nice bootstrapping. I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

PR #41, opened 2016-02-25
Done

pajamaw:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+      if user && user.authenticate(params[:password])
+        session[:id] = user.id
+        redirect "/tweets" 
+      else
+        redirect "/"
+      end
+    end
+  end
+
+  get "/tweets" do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+      session.clear
+      redirect to "/login"
+    else
+      redirect "/"
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+    else
+      @tweet.update(content: params[:content])
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete "/tweets/:id/delete" do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:id]
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/views/layout.erb
```diff
@@ -0,0 +1,38 @@
...
+     B"  ._L_yQ&Qy&&QQBQQDlL;S; (Q
+    B^   /DQQQBQQQQQQBQBDQQD1:   "B
+   Q~   ,GQBDQDQQQQQQBBQQDQ&EC    7B
+   D    lQQQQQBQBQQQQQQQQBBID('    {
+   E   .iDDBQQBQQQQBBBQBQQDQ1!:    (
+   3   .(BQBBQBBQQQQBBQBBBDBS'     }                     I'm
+   Q   '~~<MMBQQBBQBP"""<BD1EL.    Q                   dreaming
+   Q    LuL `7QBBB?../LL;.!JDL     Q                    of a
+  "  ,  ,__&L.DBBEEL1L_,'EylEE!  ,U                     white
+  1 :s&yyQQQ6-DQBBQ&QQQQQQQQJY!  lyB1Q                Christmas.
```
randallreedjr: Isn't that a Bing Crosby song?

app/models/user.rb
```diff
@@ -0,0 +1,7 @@
+class User < ActiveRecord::Base
+  include Slugifiable
+  extend Slugifiable 
+  validates_presence_of :username, :password
```
randallreedjr: Very nice. What about...uniqueness?

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      erb :'tweets/create_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  post "/tweets" do 
+    @tweet = Tweet.new(params[:tweet])
+    @tweet.user_id = session[:id]
```

``` ruby
```

or alternatively

```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+    if !logged_in?
+      erb :'users/create_user'
+    else
+      redirect "/tweets"
+    end
+  end
+
+  post "/signup" do
+    user = User.new(username: params[:username], password: params[:password], password_confirmation: params[:password_confirmation], email: params[:email])
+    if params[:username].empty? || params[:password].empty? || params[:email].empty?
```
randallreedjr: These checks should already be covered by your User model validators.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+
+  get "/users/:slug" do
+    @user = User.find_by_slug(params[:slug])
+    erb :'users/show'
+  end
+
+  get "/tweets/:id/edit" do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:id]
```

app/controllers/application_controller.rb
pajamaw: Ah I didn't think about that concerning the session[:id]. This is great 
thanks for the feedback Randall! I'll go through this and the rest of 
these and begin refactoring tomorrow.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+      redirect "/login"
+    end
+  end
+
+  patch "/tweets/:id" do
+    @tweet = Tweet.find_by_id(params[:id])
+    if params[:content].empty?
+       redirect "/tweets/#{@tweet.id}/edit"
+    else
+      @tweet.update(content: params[:content])
```

randallreedjr:
@pajamaw Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

db/schema.rb
```diff
@@ -0,0 +1,31 @@
...
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
+
+  create_table "users", force: :cascade do |t|
+    t.string   "username"
+    t.string   "email"
+    t.string   "password_digest"
+    t.datetime "created_at"
+    t.datetime "updated_at"
```
randallreedjr: Nice timestamps :sunglasses:

PR #47, opened 2016-03-05
Project submitted

davdkm:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
+class TweetsController < ApplicationController
```
randallreedjr: Great job splitting your actions between multiple controllers!

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    if logged_in?
+      redirect to '/tweets'
+    end
+    erb :'/users/create_user'
+  end
+
+  post '/signup' do
+    if logged_in?
+      redirect '/tweets'
+    elsif params.values.include?('')
```
randallreedjr: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,32 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,32 @@ class ApplicationController < Sinatra::Base
...
+      @user=current_user
+      erb :index
+    else
+      redirect '/'
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+  end
+
+  post '/tweets' do
+    if logged_in?
+      if params["content"].empty?
+        redirect '/tweets/new'
+      end
+      @tweet=Tweet.new(content: params["content"])
+      @tweet.user_id=session[:id]
```
randallreedjr: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```

or alternatively

```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+    else
+      redirect "/login"
+    end
+  end
+
+  get '/tweets/:id' do
+
+    if logged_in?
+      @tweet=Tweet.find(params[:id])
+      # @user=User.find(@tweet.user_id)
```
randallreedjr: Shouldn't you verify a tweet belongs to the `current_user` before allowing them to edit? You could use something like
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+      erb :'tweets/show_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do
+
+    if logged_in?
+      @tweet=Tweet.find(params[:id])
```
randallreedjr: Shouldn't you verify a tweet belongs to the `current_user` before allowing them to edit?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+    else
+      redirect "/login"
+    end
+
+  end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+      @tweet=Tweet.find(params[:id])
+      if @tweet.user_id==session[:id]
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

randallreedjr:
@davdkm Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #49, opened 2016-03-06
Done

Brad1974:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+  
+  get '/tweets/new' do
+    if !session[:user_id]
+      redirect to :'login'
+    else
+      erb :'tweets/create_tweet'
+    end
+  end
+
+  get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+   if !session[:user_id]
+      erb :'users/signup'
+    else
+      redirect "/tweets"
+    end
+  end
+
+  post '/signup' do
+
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+
+  helpers do
+  
+
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+      redirect to '/login'
+    end
+  end
+
+
+  helpers do
+  
+
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+    end
+  end
+
+  post '/tweets' do
+
+    if params[:content] == ""
+      redirect to '/tweets/new'
+    else
+      @tweet = Tweet.new(:content => params[:content])
+      @tweet.user_id = session[:user_id]
```

``` ruby
```

or alternatively

```
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  get '/tweets/:id' do 
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+    else 
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do 
+
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
randallreedjr: I don't see a lot of students checking the tweet's user id. Good job! Also, your `current_user` method might be useful here.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  post '/tweets/:id/delete' do 
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
+      binding
+      if @tweet.user_id == session[:user_id]
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  post '/tweets/:id/delete' do 
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
+      binding
```
randallreedjr: Tsk tsk!

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+  end
+
+
+  get '/users/:id' do 
+    @user = User.find_by(id: params[:id])
+    erb :'users/show'
+  end
+
+  get '/signup' do
+   if !session[:user_id]
```
randallreedjr: You can use your `logged_in?` helper method here, which is preferable to referring to the session object directly.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,166 @@
...
+  end
+  
+  patch '/tweets/:id' do
+
+    if params[:content] == ""
+      redirect "/tweets/#{params[:id]}/edit"
+    else
+      @tweet=Tweet.find(params[:id])
+      @tweet.content = (params[:content])
+      @tweet.save
```

randallreedjr:
@Brad1974 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #55, opened 2016-03-17
Done

bhollan:


Gemfile
```diff
@@ -17,4 +17,4 @@ group :test do
   gem 'capybara'
   gem 'rack-test'
   gem 'database_cleaner', git: 'https://github.com/bmabey/database_cleaner.git'
-end
\ No newline at end of file
+end
```
randallreedjr: Nice catch!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,94 @@
+require './config/environment'
+
+class TweetsController < Sinatra::Base
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+    if is_logged_in
+      # erb :'/tweets/index'
+      redirect to '/tweets'
+    else
+      erb :'users/create_user', layout: false
+    end
+  end
+
+  post '/signup' do
+    @user = User.create(username: params[:username], email: params[:email], password: params[:password])
```
randallreedjr: Great job using ActiveRecord model validations! The usual pattern for this is to combine `User.new` and `if @user.save?` instead of `User.create` and checking the ID.

app/models/helpers/helper.rb
```diff
@@ -0,0 +1,9 @@
+module Helper
+  def current_user
+    @current_user ||= User.find_by_id(session[:id])
```
randallreedjr: Woh! Very impressed.

app/models/helpers/helper.rb
```diff
@@ -0,0 +1,9 @@
...
+  end
+
+  def is_logged_in
+    current_user != nil
```
randallreedjr: This is good. You could also use `!!current_user`.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+      @user = current_user
+      erb :'tweets/create_tweet'
+    end
+  end
+
+  post '/tweets' do
+    if !is_logged_in
+      redirect to '/login'
+    else
+      @tweets = Tweet.create(user_id: current_user.id, content: params[:content])
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    erb :'tweets/tweets'
+  end
+
+  get '/tweets/:id/edit' do
+    if !is_logged_in
+      redirect to '/login'
+    end
+    @tweet = Tweet.find(params[:id])
+    #redirect non-tweet owning users to /tweets
+    if current_user.id != @tweet.user_id
```
randallreedjr: I don't see a lot of students checking the tweet's user id. Good job!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+    erb :'tweets/edit_tweet'
+  end
+
+  patch '/tweets/:id' do
+    if params[:content].empty?
+      redirect to "/tweets/#{params[:id]}/edit"
+    end
+    @tweet = Tweet.find(params[:id])
+    @tweet.update(content: params[:content])
```
randallreedjr: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a POST request than going through an HTML form...

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    redirect to "/tweets/#{params[:id]}"
+  end
+
+  delete '/tweets/:id/delete' do
+    if !is_logged_in
+      redirect to '/login'
+    end
+    @tweet = Tweet.find(params[:id])
+    if @tweet.user_id != current_user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

randallreedjr:
@bhollan Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

app/views/layout.erb
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Fwitter; not a copyright violation at all.</title>
```
randallreedjr: Haha

PR #57, opened 2016-03-17
Done

drywaterr24:


app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,55 @@
+class SessionsController < ApplicationController
```
randallreedjr: Great job splitting your actions among multiple controllers!

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  post '/login' do
+    binding.pry
+    @user = User.find_by(username: params[:username])
+      session[:id] = @user.id
+      erb 'tweets/tweets'
+  end
+
+  post '/signup' do
+    if params[:username].empty? || params[:email].empty? || params[:password].empty?
```
randallreedjr: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    if  @user
+      erb :'tweets/show_tweet'
+    end
+  end
+
+  get '/tweets/new' do
+    if logged_in?
+      @user = User.find(session[:id])
```
randallreedjr: Looks like you're missing a `current_user` helper method in your application controller. It is preferable to use that method both because it is more expressive and can save some session/database lookups through Memoization ([read more](http://gavinmiller.io/2013/basics-of-ruby-memoization/) and [here](http://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/)).

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,32 @@ class ApplicationController < Sinatra::Base
...
   end
 
-end
\ No newline at end of file
+
+
+  helpers do
+
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+    if params[:content] != @old_tweet && !params[:content].empty?
+      @old_tweet.content = params[:content]
+    end
+    @old_tweet.save
+  end
+
+  post '/tweets' do
+    @user = User.find(session[:id])
+    if !params[:content].empty?
+      @user.tweets << Tweet.create(content: params[:content])
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      erb :'tweets/show_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
```
randallreedjr: Shouldn't you verify a tweet belongs to the `current_user` (not just that any user is logged in) before allowing them to edit?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+    @user = User.find_by(username: params[:username])
+      session[:id] = @user.id
+      redirect to '/tweets'
+  end
+
+
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
+    if session[:id] == @tweet.user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

randallreedjr:
@drywaterr24 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #65, opened 2016-03-23
Project submitted

fiveinfinity:


app/models/user.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def slug
+    username.downcase.gsub(" ","-")
+  end
+
+  def self.find_by_slug(slug)
+    User.all.find{|user| user.slug == slug}
+  end
+
+  def self.current_user(session)
```
randallreedjr: Implementing `current_user` and `is_logged_in` as controller helper methods would make them easier to invoke and would allow you to avoid passing the `session` object as a parameter.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
+class TweetController < ApplicationController
```
randallreedjr: Great job splitting your actions among multiple controllers!

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  get '/signup' do
+    if User.is_logged_in?(session)
+      redirect to '/tweets'
+    else
+      erb :'users/create_user'
+    end
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:password] == "" || params[:email] == ""
```
randallreedjr: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/models/user.rb
```diff
@@ -0,0 +1,21 @@
...
+    @user = User.find_by_id(session["user_id"])
+    @user
+  end
+
+  def self.is_logged_in?(session)
+    !!session["user_id"]
```
randallreedjr: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      redirect to "/login"
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] == ""
+      redirect to "/tweets/new", locals: {message: "Your tweet was empty."}
+    else
+      @user = User.find_by_id(session[:user_id])
+      @tweet = Tweet.create(:content => params[:content], :user_id => @user.id)
```
randallreedjr: Another way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+      erb :'tweets/show_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    @tweet = Tweet.find_by(params[:id])
+    if User.is_logged_in?(session)
```
randallreedjr: Shouldn't you verify a tweet belongs to the `current_user` (not just that any user is logged in) before allowing them to edit?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find_by(params[:id])
+    if params[:content].empty?
+      redirect "/tweets/#{@tweet.id}/edit"
+    else
+      @tweet.content = params[:content]
+      @tweet.save
```
randallreedjr: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a POST request than going through an HTML form...

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      redirect "/tweets", locals: {message: "Tweet was successfully edited."}
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    if User.is_logged_in?(session)
+      @tweet = Tweet.find_by_id(params[:id])
+      if User.current_user(session).id == @tweet.user_id
+        @tweet.delete
```
randallreedjr: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

randallreedjr:
@fiveinfinity Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #68, opened 2016-03-27
Done

jamesnvk:


app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,26 @@
+class SessionsController < ApplicationController
+
+  get '/login' do
+    if !!session[:id]
```
randallreedjr: You could use the `current_user` and `logged_in?` helper methods here, which is preferable to referring to the session object directly.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
+class TweetsController < ApplicationController
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  get '/signup' do
+    if !!session[:id]
+      redirect to "/tweets"
+    else 
+      erb :'users/create_user'
+    end
+  end
+
+  post '/signup' do
+    if params.has_value?("")
```
randallreedjr: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+  end
+
+  post '/tweets' do
+    if params.has_value?("")
+      redirect to "/tweets/new"
+    else
+      @user = User.find(session[:id])
+      @tweet = Tweet.create(params)
+      @user.tweets << @tweet
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      erb :'tweets/show_tweet'
+    else
+      redirect to "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do 
+    if !!session[:id]
+      @user = User.find(session[:id])
+      @tweet = Tweet.find_by(params[:id])
```
randallreedjr: Shouldn't you verify a tweet belongs to the `current_user` (not just that any user is logged in) before allowing them to edit?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find_by(params[:id])
+    if params.has_value?("")
+      redirect to "/tweets/#{@tweet.id}/edit"
+    else
+      @tweet.content = params[:content]
+      @tweet.save
```
randallreedjr: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a POST request than going through an HTML form...

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      @tweet.content = params[:content]
+      @tweet.save 
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    if !!session[:id]
+      @tweet = Tweet.find_by(params[:id])
+      @tweet.destroy
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

randallreedjr:
@jamesnvk Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #69, opened 2016-03-27
Done

gormanjp:


app/helpers.rb
```diff
@@ -0,0 +1,12 @@
+class Helpers
+#helpers
+  def self.current_user(user)
```
randallreedjr: Using the variable `user` to refer to a `session` is a little confusing. The session can store different data than the `User` model. Also, the session should be directly available in the helper methods, without having to pass it as a parameter.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,172 @@
...
+  end
+
+  #Log Out
+  get '/logout' do 
+  	session.clear
+  	redirect '/login'
+  end
+
+  #Home Page - tweets feed
+  get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,172 @@
...
+
+
+
+
+
+
+
+
+
+
```
randallreedjr: A little extra space here, eh?

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,172 @@
...
+  	if Helpers.is_logged_in?(session)
+  		redirect '/tweets'
+  	else
+  		erb :'users/signup'
+  	end
+  end
+
+  post '/signup' do 
+  	@user = User.create(:username => params[:username], :email => params[:email], :password => params[:password])
+  	@user.save
```
randallreedjr: No input validation?

app/helpers.rb
```diff
@@ -0,0 +1,12 @@
...
+  	@user = User.find_by_id(user[:id])
```
randallreedjr: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/helpers.rb
```diff
@@ -0,0 +1,12 @@
...
+  	@user
+  end
+
+  def self.is_logged_in?(user)
+  	!!self.current_user(user)
```
randallreedjr: 👍

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,172 @@
...
+  get '/tweets/new' do 
+  	if Helpers.is_logged_in?(session)
+	  	@user = User.find_by_id(session[:id])
+  		erb :'tweets/new'
+  	else
+  		redirect '/login'
+  	end
+  end
+
+  post '/tweets/new' do
```
randallreedjr: Typically, you would post to `/tweets` instead of `/tweets/new`.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,172 @@
...
+  	else
+  		redirect '/login'
+  	end
+  end
+
+  post '/tweets/new' do 
+  	if params[:content] != ""
+  		@tweet = Tweet.create(:content => params[:content])
+  		@user = User.find_by_id(session[:id])
+  		@user.tweets << @tweet
```

``` ruby
```

or alternatively

```
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,172 @@
...
+  		@tweet = Tweet.find_by(:id => params[:id])
+  		erb :'tweets/show'
+  	else
+  		redirect '/login'
+  	end
+  end
+
+  #Edit
+  get '/tweets/:id/edit' do 
+  	if Helpers.is_logged_in?(session)
```
randallreedjr: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,172 @@
...
+  		erb :'tweets/edit'
+  	else
+  		redirect '/login'
+  	end
+  end
+
+  patch '/tweets/:id/edit' do
+  	@tweet = Tweet.find_by(:id => params[:id])
+    @user = User.find_by_id(session[:id])
+    redirect "/tweets" if @user.id != @tweet.user_id
```
randallreedjr: :

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,172 @@
...
+  	else
+  		redirect "tweets/#{@tweet.id}/edit"
+  	end
+  end
+
+  #Delete
+  delete '/tweets/:id/delete' do 
+  	@user = User.find_by_id(session[:id])
+  	@tweet = Tweet.find_by(:id => params[:id])
+  	Tweet.delete(params[:id]) if @user.id == @tweet.user_id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

randallreedjr:
@gormanjp Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #79, opened 2016-04-03
done

lawrend:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,148 @@ class ApplicationController < Sinatra::Base
...
+    if !@tweet.content.empty?
+      @tweet.save
+      redirect "/tweets/#{@tweet.id}"
+    else
+      redirect "/tweets/new"
+    end
+  end
+
+### SHOW ALL TWEETS ###
+  get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,148 @@ class ApplicationController < Sinatra::Base
...
+    if Helpers.is_logged_in?(session)
+      redirect "/tweets"
+    else
+      erb :'/users/create_user'
+    end
+  end
+
+  post '/signup' do 
+    @user = User.new(username: params["username"], email: params["email"], password: params["password"])
+    if !@user.username.empty? && !@user.email.empty? && @user.password != nil
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,11 @@
+class Helpers
+  
+
+    def self.current_user(session)
+      @user = User.find(session[:user_session_id])
```
randallreedjr: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,11 @@
...
+    end
+
+    def self.is_logged_in?(session)
+      !!session[:user_session_id]
```
randallreedjr: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,11 @@
+class Helpers
+  
+
+    def self.current_user(session)
+      @user = User.find(session[:user_session_id])
+    end
+
+    def self.is_logged_in?(session)
```
randallreedjr: I believe your helper methods should have access to the `session` object without having to pass it in as a parameter.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,148 @@ class ApplicationController < Sinatra::Base
...
+### CREATE NEW TWEET ###
+  get '/tweets/new' do
+    if Helpers.is_logged_in?(session)
+      erb :'/tweets/create_tweet' 
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets/new' do
```
randallreedjr: Typically, tweet creation would be posted to `/tweets` instead of `/tweets/new`.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,148 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/new' do
+    if Helpers.is_logged_in?(session)
+      erb :'/tweets/create_tweet' 
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets/new' do 
+    @tweet = Tweet.new(:content => params["content"], :user_id => params["user_id"])
```
randallreedjr: `user_id` should not be in the params. A malicious user could use that to spoof another user's account and pose as them. Instead, use the session's user, `current_user`.

Another way to create the tweet and associate it to the user is to make use of the `has_many` relationship on User:
```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,148 @@ class ApplicationController < Sinatra::Base
...
+        redirect "/login"
+      end
+    end
+
+### EDIT SPECIFIC TWEETS ###
+  get '/tweets/:id/edit' do 
+    if Helpers.is_logged_in?(session)
+      @inquiring_user = Helpers.current_user(session)
+      @tweet = Tweet.find(params[:id])
+        if @inquiring_user.id == @tweet.user_id
```
randallreedjr: I don't see a lot of students checking the tweet's user id. Good job!

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,148 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+  patch '/tweets/:id' do 
+    if params["content"].empty?
+      redirect "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find(params[:id])
+      @tweet.content = params["content"]
+      @tweet.save
```
randallreedjr: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,148 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+  ### DELETE SPECIFIC TWEET
+  delete '/tweets/:id/delete' do 
+    if Helpers.is_logged_in?(session)
+      @user = Helpers.current_user(session)
+      @tweet = Tweet.find(params[:id])
+      if @user.id == @tweet.id
+        @tweet.destroy
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/views/layout.erb
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+  <meta charset="UTF-8">
+    <title>Fwitter</title>
+    <link rel="stylesheet" href="stylesheets/bootstrap.css" type="text/css">
```
randallreedjr: Above and beyond!

randallreedjr:
@lawrend Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

app/controllers/application_controller.rb
lawrend: I think the tests are set up so that they only pass if you are running everything through one controller.

lawrend: That first reply was supposed to be in response to this comment in particular.  (one controller to rule them all, and one controller to pass the tests)

app/controllers/application_controller.rb
lawrend: Yeah I had validates_presence_of set up in the models.  My mistake was that my method didn't check correctly (validates_presence_of default validates at save, not create) so I should have written something more like 
	if @user.save
		<continued on with function>

I actually did that in my sinatra project after I figured out how to validate correctly.

app/helpers/helpers.rb
lawrend: Ah, that makes so much sense.  

I made the is_logged_in? and current_user methods for a lab in Sinatra waaaaaay before and the way I got it to work just stayed the way I worked it.  

So now my helper methods in my sinatra project look all goofy too.  But this is way cleaner (and I see the later comments about access to session).

Thanks for this!

app/controllers/application_controller.rb
lawrend: The way I managed this (and I'm sure this is weird) is that in the view erb :'/tweets/create_tweet', I used erb to call Helpers.is_logged_in? as a conditional and if the user was logged in, I set a local variable "user" to Helpers.current_user. (so user=Helpers.current_user)

My concern was exactly what you mentioned, a user somehow posting a tweet as someone else, so I set the user_id param as a hidden input with a name of "user_id" and a value of <%=user.id%>, so that when I created the tweet in the controller's post '/tweets/new' route the tweet would belong to whomever was logged in.  Make sense?

It feels like a hidden input can still be messed with though, yeah?  Maybe by inspecting the element or something?  I honestly don't know but since it is being sent as an HTML request it's out there somehow and is less secure than taking care of it server side in the controller.

And I didn't even consider using the has_many relationship.  My brain just isn't there yet :) 

I don't think I even knew you could call create like that.  That's awesome.

app/controllers/application_controller.rb
lawrend: Ah, that's good to know.  I think i've been operating under the assumption that if I've already checked id's at the outdoor festival that is the form that I didn't need to check them again when those data were coming back inside, and that's because I didn't know there were other ways to submit a PATCH/POST request except for HTML.  

I want to know how to submit w/o HTML and I also don't want to know.
Because if I know then I want to try.  
And if I try I want it to work.  
And if it works I want it to have hilarious consequences.

Like, I'd have liked to have had a part in this...

http://imgur.com/a/XE92y <http://imgur.com/a/XE92y>

app/controllers/application_controller.rb
lawrend: Nice!  There it is again, I didn't realize I could call something like find_by on current_user.tweets.

It's geeky AF to say, but ActiveRecord is pretty awesome sometimes.

app/views/layout.erb
lawrend: Just for fun I slapped Bootstrap onto my final Sinatra project, just for the index page I told myself...

...and then I couldn't stop.  I maybe spent time over 2 or 3 days making every page look decent.  But once one page looked good the others looked horrible.

I figure it's that much closer to something I can show a prospective employer, and my girlfriend was way more impressed with my project once it looked like every other Bootstrap site.

lawrend:
Thanks again for all your attention and comments.  You looked over my ttt w AI project a month back and I was thankful then too.  

And to think, I knew zero Sinatra then.

See you in a month when you get to see a bunch more silliness from me, but this time in Rails!  

Doug

app/controllers/application_controller.rb
randallreedjr: Your hypothesis was correct. Try it in Chrome! You can right click a form field, click `inspect`. Then find the hidden field, right click, and select `Edit as HTML`. Additionally, this could be manipulated with a utility such as `curl` or an app like [Postman](https://www.getpostman.com/).

PR #80, opened 2016-04-04
Project submitted

rocketNeck:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+      erb :"/tweets/show_tweet"
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:id]
```
randallreedjr: `current_user` here too.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+    if @user
+      session[:id] = @user.id
+      redirect to '/tweets'
+    else
+      redirect to '/signup'
+    end
+  end
+  ##################################################
+
+  get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+  get '/signup' do
+    if logged_in?
+      redirect to '/tweets'
+    else
+      erb :'/users/create_user'
+    end
+  end
+
+  post '/signup' do
+    if params.any? { |_k, v| v.empty? }
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+    redirect to '/tweets'
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
+    @tweet.delete if current_user.id == @tweet.user_id
+    redirect to '/tweets'
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
\ No newline at end of file
+  get '/tweets/new' do
+    if logged_in?
+      erb :'/tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/new' do
```
randallreedjr: Typically, tweet creation would be posted to `/tweets` instead of `/tweets/new`.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+    if params[:content] == ''
+      redirect '/tweets/new'
+    else
+      @tweet = Tweet.create(params)
+      @tweet.user = current_user
+      @tweet.save
```

``` ruby
```

or alternatively

```
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+
+  patch '/tweets/:id/edit' do
+    if !logged_in?
+      redirect '/login'
+    elsif params[:content] == ''
+      redirect to "/tweets/#{@tweet.id}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params[:content]
+      @tweet.save
```
randallreedjr: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,149 @@
...
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
+    @tweet.delete if current_user.id == @tweet.user_id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

randallreedjr:
@rocketNeck Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

app/controllers/application_controller.rb
rocketNeck: I think I am foggy about the conventions here. How do we decide on this say for instance if there are multiple forms on one view?

app/controllers/application_controller.rb
rocketNeck: I think here I tried to use current_user  rather than !!current_user and it didn't work so I fell back to !!session. Thank you for correction.

app/controllers/application_controller.rb
rocketNeck: Thank you! I had seen this operator and had not found what it was for yet.

app/controllers/application_controller.rb
rocketNeck: i had the validates_presence_of :username in the user model all along. for some reason it kept getting  2 failed tests. I tried this and it worked. I think I would need to find out why validates pressance of doesn't work.

app/controllers/application_controller.rb
rocketNeck: trying to make the changes.  when i use `@current_user ||= User.find(session[:id] if session[:id])` i get an error  unexpected modifier_if,  
when i just use `User.find(session[:id])` it works

app/controllers/application_controller.rb
rocketNeck: when i use `!!current_user`  95% of the tests fail

app/controllers/application_controller.rb
randallreedjr: @rocketNeck The destination of a form is based on the target of the form, not the source. For example:

> Creating a new tweet -> POST to tweets
> Update an existing tweet -> PATCH to tweets/:id
> Delete a tweet -> DELETE to tweets/:id

This is referred to a 'RESTful routes' and will be covered later in the Learn curriculum. You can also read more [here](http://guides.rubyonrails.org/v2.3.11/routing.html).

app/controllers/application_controller.rb
randallreedjr: @rocketNeck Hm, this may be due to operator precedence. Try using parentheses to force the intended order, e.g. `(@current_user ||= User.find(session[:id]) if session[:id])`.

app/controllers/application_controller.rb
randallreedjr: Hmm, that is unexpected. If you upload your updated code, I will try to investigate this.

PR #81, opened 2016-04-04
Project submitted

jakeekaj:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,91 @@
+class TweetsController < ApplicationController
+
+
+  get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,86 @@ class ApplicationController < Sinatra::Base
...
+  get "/signup" do
+  	if logged_in?
+      redirect "/tweets"
+    else
+      erb :signup
+    end
+  end
+
+  post "/signup" do
+    if params[:username] == "" || params[:password] == "" || params[:email] == ""
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,86 @@ class ApplicationController < Sinatra::Base
...
+    erb :'/users/show'
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,86 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    erb :'/users/show'
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    erb :'tweets/show'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if !logged_in?
+      redirect "/login"
+    else
+    @tweet = Tweet.find(params[:id])
+      if current_user.tweets.include?(@tweet)
```
randallreedjr: I don't see a lot of students checking the tweet's user id. Good job! 💯

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
+    if params[:content] == "" 
+      redirect '/tweets/' + @tweet.id.to_s + '/edit'
+    else
+    @tweet.content = params[:content]
+    @tweet.save
```
randallreedjr: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    redirect '/tweets/' + @tweet.id.to_s
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    if current_user.tweets.include?(Tweet.find(params[:id]))
```
randallreedjr: Beautifully done. 👍 👍

randallreedjr:
@jakeekaj Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

app/controllers/application_controller.rb
jakeekaj: Thank you Randall! I'll look into Memoization. It's the first time I've heard about it. However, I remember using | |= in the past.

app/controllers/application_controller.rb
jakeekaj: Thank you. This is noted. :)

app/controllers/tweets_controller.rb
jakeekaj: Dear Randall,

Could you elaborate a bit on these "other ways to submit a PATCH/POST request"?

Thank you!

randallreedjr: Hi @jakeekaj. You can use utilities like `curl` or the app [Postman](https://www.getpostman.com/) to submit any http request. Try running `curl http://www.google.com` from your terminal!

PR #184, opened 2016-08-11
refactored

gj:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
+    end
+  end
+end
```
mendelB: Great job splitting your actions among multiple controllers!

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    if !logged_in?
+      erb :'users/create_user'
+    else
+      redirect '/tweets'
+    end
+  end
+
+  post '/signup' do
+    user = User.new(username: params[:username], email: params[:email], password: params[:password])
+    if !user.username.empty? && !user.email.empty? && user.save
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+    erb :index
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
```
mendelB: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,9 @@
+Welcome, <%= current_user.username %>
```
mendelB: Here's an example of where you call the `current_user` a second time causing a second db query, it would be doubly useful to avoid that in this instance since we try to avoid making db calls from the views.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
 
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
```
mendelB: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/views/tweets/create_tweet.erb
```diff
@@ -0,0 +1,5 @@
+<h2>tweet something</h2>
+<form action="/tweets" method="POST">
```
mendelB: Nice job posting to `'/tweets'` here  and not `'/tweets/new'`!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content].empty?
+      redirect '/tweets/new'
+    else
+      @tweet = Tweet.create(content: params[:content], user_id: current_user.id)
```
mendelB: Another way to do this is to make use of the `has_many` relationship on User:
```
```
If you do go with your approach, a slightly more elegant form would be: 
`@tweet = Tweet.create(content: params[:content], user: current_user)`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+        erb :'tweets/edit_tweet'
+      else
+        redirect '/tweets'
+      end
+    else
+      redirect '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      @tweet.content = params[:content]
+      @tweet.save
+      redirect "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do 
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
+      if @tweet.user_id == current_user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

mendelB:
@gj Really well done!!
I've added a few notes throughout the code with some refactoring tips.
It'd be a good idea to go through and research the concepts and appropriately refactor. 
This will give you good refactoring practice, as well as prove useful in your assessment project.

Good luck!!

app/controllers/users_controller.rb
gj: Hey @mendelB, that Learn lesson you linked doesn't actually seem to mention anything about Active Record validations. It says:
> We'll validate some of the attributes of our user by writing code that makes sure no one can sign up without inputing their name, email and password. More on this later.

and then later adds:
> You'll see that you're passing a number of tests, including tests like these:
```ruby
User
is invalid without a name
is invalid without a email
is invalid without an password
```
Let's think about the concept of validations...

but it never actually seems to elaborate on validations, Active Record or otherwise.

app/controllers/tweets_controller.rb
gj: Ask about `find_by` vs `find`

app/controllers/tweets_controller.rb
gj: Ask about this

app/controllers/application_controller.rb
gj: Ask about this

app/controllers/application_controller.rb
gj: @mendelB -- just to verify, that @current_user instance variable will be visible in all the controllers that inherit from ApplicationController and (therefore) all views, correct?

app/controllers/tweets_controller.rb
gj: @mendelB shouldn't `redirect "/tweets/#{tweet.id}"` be `redirect "/tweets/#{params[:id]}"`?

PR #186, opened 2016-08-16
Done

jwolfe890:


app/controllers/users_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+ get '/signup' do 
+    if !is_logged_in?
+      erb :'users/create_user'
+    else 
+      redirect "/tweets"
+    end 
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      redirect '/login'
+    else 
+      erb :'/tweets/create_tweet'
+    end 
+  end
+
+  post '/tweets' do 
+    if !@params["content"].empty?
+      @tweet = Tweet.create(:content => @params["content"])
+      current_user.tweets << @tweet
```

``` ruby
```

or alternatively

```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      erb :'/tweets/show_tweet' 
+    else
+      redirect to '/login'
+    end
+  end
+
+    get '/tweets/:id/edit' do 
+      if is_logged_in?
+        @tweet = Tweet.find_by_id(params[:id])
+        if @tweet.user_id == session[:user_id]
```
mendelB: I don't see a lot of students checking the tweet's user id. Good job!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+        end
+      else 
+        redirect '/login'
+    end
+    end 
+
+  patch '/tweets/:id' do
+   if params[:content] == ""
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+    else  
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params["content"]
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+  end 
+end 
+
+  delete '/tweets/:id/delete' do 
+    if is_logged_in?
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

mendelB:
Hey @jwolfe890 This is awesome!

I left a couple of refactoring comments/tips. I'd highly recommend implementing them and @ me when you're done so I can do another quick review.

Looking forward to talking soon!

PR #187, opened 2016-08-16
Project submitted

achasveachas:


app/controllers/users_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  set :views, 'app/views'
+
+# Displayse the "Sign Up" form and creates a new user
+  get '/signup' do
+    redirect '/tweets' if logged_in?
+    erb :'users/create'
+  end
+
+  post '/signup' do
+    redirect '/signup' if params['username'] == "" || params['email'] == "" || params['password'] == ""
```
Also, since the params hash keys are named the same as the actual attributes, you can get away with just running `user = User.create(params)` or if incorporating Active Record validations: `user = User.new(params)`

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,39 @@
...
+    erb :index
+  end
+
+  helpers do
+
+    def logged_in?
+      !!session[:user]
+    end
+
+    def current_user
```
mendelB: Generally, the `current_user ` helper method would query the database for the user with the same id as the `session[:id]` and would return that instance. e.g.
``` ruby
def current_user
  User.find_by(id: session[:id])
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    redirect '/login' if !logged_in?
+    erb :'/tweets/new'
+  end
+
+  post '/tweets' do
+    redirect '/login' if !logged_in?
+    redirect '/tweets/new' if params[:tweet][:content].empty?
+    @tweet = Tweet.create(params[:tweet])
+    @tweet.user_id = User.find_by(username: current_user).id
+    @tweet.save
```

``` ruby
```

or alternatively

```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    @tweet.delete if @user.id == @tweet.user_id
+    redirect '/tweets'
+  end
+
+# Edits an individual tweet
+  get '/tweets/:id/edit' do
+    redirect '/login' if !logged_in?
+    @tweet = Tweet.find(params[:id])
+    @author =
+    redirect '/tweets' if @tweet.user.username != current_user
```
mendelB: I don't see a lot of students checking the tweet's user id. Good job!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    erb :'/tweets/edit'
+  end
+
+  post '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
+    redirect "/tweets/#{@tweet.id}/edit" if params[:tweet][:content].empty?
+    @tweet.content = params[:tweet][:content]
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+# Shows an individual tweet
+  get '/tweets/:id' do
+    redirect '/login' if !logged_in?
+    @tweet = Tweet.find(params[:id])
+    erb :'/tweets/show'
+  end
+
+  #Deletes a tweet.
+  delete '/tweets/:id/delete' do
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

mendelB:
Hi @achasveachas @CFaught this is awesome!

I left a bunch of comments and refactoring tips throughout the code, hope you find them helpful!

I'd highly recommend refactoring the project based on those notes. @ me when you're done and I can give this one more quick scan.

Good luck!

app/controllers/users_controller.rb
achasveachas: The lesson you linked to doesn't mention that method of validating.
By coincidence, I was just part of session with Luke where he mentioned it so I googled it and learned something new :)

app/controllers/application_controller.rb
achasveachas: That makes sense!
Was a bitch to refactor, had to hunt down every time current_user was called plus a whole bunch of other bugs that came up, but yeah, i got it in the end :)

app/controllers/tweets_controller.rb
achasveachas: I like the second one, has a nice elegance to it...

app/controllers/tweets_controller.rb
achasveachas: Makes sense...

app/controllers/tweets_controller.rb
achasveachas: @mendelB small problem with your suggestion:
If `tweet` is false you redirect to `"/tweets/#{tweet.id}"`but if `tweet` is false you will probably get a "no method id for false:Class" error.
You probably meant to redirect to `"/tweets/#{params[:id]}"`

CFaught: This makes sense and is more elegant. I had to make some slight changes to make it work:
`else
  redirect "/tweets/#{ params[:id] }" 
end`

Thanks @mendelB

CFaught:
@mendelB, We've gone through and refactored a bit if you want to look it over again. Thanks for the tips!

app/controllers/tweets_controller.rb
mendelB: Yeah you're both right!
That definitely does not make sense - thanks for pointing that out!

mendelB:
This is perfect!

PR #188, opened 2016-08-16
Done

bhabig:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
+require 'pry'
+
+class TweetsController < ApplicationController
+  get '/tweets' do
+    if !!session[:user_id]
```
mendelB: You can use your `logged_in?` helper method here, which is preferable to referring to the session object directly.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      @user = User.find(session[:user_id])
```
mendelB: Your `current_user` method could be useful here.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
+    end
+  end 
+end
```
mendelB: Great job splitting your actions among multiple controllers!

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+  end
+
+  get "/users/:slug" do
+    @user = User.find_by_slug(params[:slug])
+    redirect "/tweets/#{user.slug}"
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:password] == "" || params[:email] ==""
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+    erb :index
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
```
mendelB: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
 
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
```
mendelB: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    if !params[:content].empty?
+      @user = User.find(session[:user_id])
+      @tweet = Tweet.create(content: params[:content])
+      @user.tweets << @tweet
```

``` ruby
```

or alternatively

```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      erb :"/tweets/show_tweet"
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if !!session[:user_id]
+      @user = User.find(session[:user_id])
+      @tweet = Tweet.find(params[:id])
```
mendelB: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      erb :"/tweets/edit_tweet"
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+  end
+
+  post "/tweets/:id/delete" do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
+      if session[:user_id] == @tweet.user_id
+        @tweet = Tweet.find(params[:id])
+        @tweet.delete
+        redirect "/tweets"
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
  redirect "/tweets/#{params[:id]}"
end
```

mendelB:
Hey @bhabig Awesome stuff!

I left a few refactoring tips and comments throughout the code. I'd highly recommend implementing these patterns in the project, and then you can @ me when you're done and I can give this another quick look through.

Best of luck! Ping me with any questions.

bhabig:
Hey Mendel I've read through your notes and they are very helpful, thank
you! most of it seems to come down to utilizing current_user more, and the
user - tweet association. i think from your code suggestions i understand
the user to tweet association better , but i definitely had a bit of a hard
time thinking of how to do that, which is why i wound up with what i did
(not knowing the difference between find and find_by was a roadblock for
sure). I am going through previous sections of the course again to get a
deeper understanding of things bc i felt more lost here than i was
comfortable with, so i will probably have a few questions for you at some
point but won't be for several days. Thanks again!

mendelB:
Hey Bob I'm so glad to hear these note were helpful!
There are definitely some concepts here that are a tad advanced and hard to really grasp but definitely worth the struggle as they'll be super helpful for the rest of the Rails journey...
Good luck with the reviewing!!

PR #192, opened 2016-08-21
Done

ebanh:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,28 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      redirect to "/tweets"
+    else
+      erb :"/index"
+    end
+  end
+
+  helpers do
+
+    def current_user
```
Lukeghenco: a great thing here is to build an instance variable for `@current_user` that way we don't have to make an activeRecord query each time we load a page if it is set. 

example: 

```ruby
@current_user ||= User.find(session[:user_id]
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,28 @@ class ApplicationController < Sinatra::Base
...
+      User.find(session[:user_id])
+    end
+
+    def logged_in?
```
Lukeghenco: we already have a method for current User lets use that here 
```ruby
!!current_user
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,68 @@
+class TweetController < ApplicationController
+
+  get '/tweets' do
```
Lukeghenco: does this page show a list of tweets?

I would imagine you would want something here that sets an instance variable for all tweets.

```ruby
@twets = Tweet.all 
``` 

?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    if logged_in?
+      erb :"/tweets/new"
+    else
+      redirect to "/login"
+    end
+  end
+
+  post '/tweets' do
+    if !params[:content].empty?
+      @tweet = current_user.tweets.create(params)
```
Lukeghenco: an instance variable is not necessary here as you will not be using @tweet in the .erb file. 

`tweet = current_user.tweets.create(params)` will be sufficient

and on line 22 you could just use 

```ruby
redirect to "/tweets/#{tweet.id}"
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      else
+        erb :"/tweets/edit"
+      end
+    else
+      redirect to "/login"
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
```
Lukeghenco: again an instance variable is not needed here.

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,53 @@
+class UserController < ApplicationController
+
+  get '/users' do
```
Lukeghenco: do we want a route to get all users?

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  get '/signup' do
+    if logged_in?
+      redirect to "/tweets"
+    else
+      erb :"/users/signup"
+    end
+  end
+
+  post '/signup' do
+    @user = User.create(params)
```
Lukeghenco: an instance variable is not needed here. Just change `@user` to be `user`

app/models/users.rb
```diff
@@ -0,0 +1,12 @@
+class User < ActiveRecord::Base
+  include Slug
+  extend ClassSlug
+
+  has_many :tweets
+
+  has_secure_password
+
+  validates_presence_of :username
```
Lukeghenco: Great use of Active Record methods here :)

app/views/tweets/index.erb
```diff
@@ -0,0 +1,12 @@
+<h3>Welcome, to Fwitter <%=current_user.username%>!</h3>
+
+<% Tweet.all.each do |tweet| %>
```
Lukeghenco: you should use that `@tweets` from the application controller I told you about here. Abstract the business logic from the view whenever possible.

PR #193, opened 2016-08-22
Done

johnsontaiwo:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,25 @@ class ApplicationController < Sinatra::Base
...
+      end
+    end
+
+    helpers do
+    def is_logged_in?
+      !!current_user
+    end
+
+    def current_user
+      @current_user ||= User.find(session[:user_id]) if session[:user_id]
```
Lukeghenco: @johnsontaiwo the `if session[:user_id]` is not needed here. it will return false if neither sides are true, just adds unecessary logic

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,77 @@
+class TweetsController < ApplicationController
+   
+   get '/tweets' do 
+    if is_logged_in?
+
```
Lukeghenco: @johnsontaiwo why is there not an instance variable here for `@tweets = Tweet.all` ?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+   if is_logged_in?
+    erb :'/tweets/create_tweet'
+   else 
+    redirect '/login'
+    end
+  end
+
+post '/tweets/new' do
+   if !params[:content].empty? 
+      @tweet = current_user.tweets.create(content: params[:content])
```
Lukeghenco: @johnsontaiwo not necessary to use an instance variable here. You can just name this `tweet` and on line 23 just use `redirect to "/tweets/#{tweet.id}"`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+    erb :'/tweets/show_tweet'
+    else
+    redirect to '/login'
+    end
+  end
+  
+  get '/tweets/:id/edit' do
+    if is_logged_in?
+      @tweet = Tweet.find(params[:id])
+      if @tweet.user_id == session[:user_id]
```
Lukeghenco: @johnsontaiwo you have a current_user method that you should be using here instead of `session[:user_id]`. `current_user.id` would be better.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+     end
+    
+  end
+
+  patch '/tweets/:id' do
+    
+     if params[:content].empty?
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
```
Lukeghenco: @johnsontaiwo again no need for instance variable here just use a method scope variable since we won't be accessing this in our erb file.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      @tweet.content = params['content']
+      @tweet.save
+       redirect "/tweets/#{@tweet.id}"
+   end
+
+end
+
+delete '/tweets/:id/delete' do
+  
+    @tweet = Tweet.find_by_id(params[:id])
```
Lukeghenco: @johnsontaiwo again no need for that instance variable again.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,56 @@
+class UsersController < ApplicationController
+
+  get '/users' do
```
Lukeghenco: do we need this route? @johnsontaiwo are we going to be needing a list of users.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    erb :'/users/index'
+  end
+ 
+ get '/signup' do
+  if  is_logged_in?
```
Lukeghenco: good using of helper method. @johnsontaiwo

app/models/users.rb
```diff
@@ -0,0 +1,14 @@
+class User < ActiveRecord::Base
+  has_many :tweets
+  has_secure_password
+
+  validates_presence_of :username
```
Lukeghenco: good use of active record method here.

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,10 @@
+Welcome, <%= current_user.username%>
+  <br>
+  <%Tweet.all.each do |tweet| %>
```
Lukeghenco: @johnsontaiwo this should use `@tweets` that you should be defining for this route in the TweetsController. Keep business logic out of the views when possible.

app/models/users.rb
johnsontaiwo: Hi Luke,
Thanks for your observations. There are somethings that were initially the
way you suggested but while testing the application via shotgun, some
things were not really working so I had to like tweak it a bit. For
instance the "if session[:user_id] when I took it out the test failed but
when I added it the test passed and some others also.
Be that as it may, I will still like to through it with you for better
understanding.
Once again, thank you so much, I really appreciate it.

Johnson.

PR #194, opened 2016-08-26
Done

KentLeong:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
+class TweeetsController < ApplicationController
+  get '/tweets/new' do
+    if !session[:user_id]
```
Lukeghenco: I would set up a helper method in the application controller to check if a user is logged_in?

examples of this would be 

```ruby
helpers do
    def logged_in?
      !!session[:user_id]
    end

    def current_user
      User.find(session[:user_id])
    end

  end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      redirect '/login'
+    else
+      erb :'/tweets/create'
+    end
+  end
+  get '/tweets' do
+    @tweets = Tweet.all
+    if !session[:user_id]
```
Lukeghenco: use the new logged_in? method here

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      redirect '/login'
+    else
+      erb :'/tweets/show'
+    end
+  end
+  get '/tweets/:id' do
+    if session[:user_id]
```
Lukeghenco: logged_in? method again.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+      erb :'tweets/tweet'
+    else
+      redirect to '/login'
+    end
+  end
+  get '/tweets/:id/edit' do
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
Lukeghenco: I would use the current_user method in the application controller that I had you add here 

```
current_user.id
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    if params[:content] == ""
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
```
Lukeghenco: we don't need to use an instance variable here on line 42 as we are not going to be using this in the view. 
I would set line 43 and 44 up using the Active Record update method 

```ruby 
tweet.update(content: params[:content]
``` 
and now we won't need to use the `.save` method anymore

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      @tweet.content = params[:content]
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
```
Lukeghenco: again this controller method does not need to have an instance method.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
Lukeghenco: update this to use the `current_user` method

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,62 @@
+
+class UsersController < ApplicationController
+
+  get '/signup' do
+    if !!session[:user_id]
```
Lukeghenco: use our new logged_in? method here

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+      redirect to '/tweets'
+    else
+      erb :'/users/signup'
+    end
+  end
+
+  get '/logout' do
+    if session[:user_id] !=nil
```
Lukeghenco: again logged_in?

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+      session.destroy
+      redirect '/login'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/login' do
+    if !!session[:user_id]
```
Lukeghenco: if logged_in? would be better here and I would have them get redirected to `/tweets` instead of a new tweet form

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+      redirect to '/tweets/new'
+    else
+      erb :'/users/login'
+    end
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    @tweets = @user.tweets
```
Lukeghenco: You will already have access to @user.tweets on the view, so we don't need to set `@tweets`. We should also check for `if logged_in?` before we set the @user variable here.

app/views/tweets/edit.erb
```diff
@@ -0,0 +1,5 @@
+<form action="/tweets/<%=@tweet.id%>" method="POST">
+  <input id="hidden" type="hidden" name="_method" value="PATCH">
```
Lukeghenco: Good job here with the hidden tag. 👍

PR #195, opened 2016-08-26
Done

kromoser:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,145 @@ class ApplicationController < Sinatra::Base
...
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "secret"
+  end
+
+  get '/' do 
+    erb :index
+  end
+
+  get '/signup' do
```
Lukeghenco: for lines 16- 137, I would separate this into 2 separate controllers: `users_controller.rb` and `tweets_controller.rb`. This will allow us to have separation of concerns, leaner controllers and more readable code.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,145 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```
Lukeghenco: I like to use an instance variable here to see if something has already been set

```ruby
@current_user ||= User.find(session[:user_id])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,145 @@ class ApplicationController < Sinatra::Base
...
+      tweet.delete
+      erb :'/tweets/deleted_tweet'
+    else
+      redirect to "/login"
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
```
Lukeghenco: you have a method that already checks for a user that has a session. 
I would just use 
```ruby
!!current_user 
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,145 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  get '/users/:slug' do 
+    @user = User.find_by_slug(params[:slug])
+    erb :'/tweets/show_tweet'
+  end
+
+  get '/tweets/new' do 
+    if logged_in?
+      @user = current_user
```
Lukeghenco: you view should have access to the current_user (@current_user variable here)

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,145 @@ class ApplicationController < Sinatra::Base
...
+      erb :'/tweets/create_tweet'
+    else
+      redirect to "/login"
+    end
+  end
+
+  post '/tweets' do 
+    if params[:content] != ""
+      @tweet = Tweet.create(content: params[:content])
```
Lukeghenco: you should be able to condense line 83-85 into one line. We also don't need to set an instance variable here since we won't be accessing it in the view.

example: 

```ruby
tweet = Tweet.create(content: params[:content], user_id: current_user.id)
```

We won't need to use .save anymore as create already does this for us.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,145 @@ class ApplicationController < Sinatra::Base
...
+      redirect to "/tweets"
+      
+    else
+      redirect to '/tweets/new'
+    end
+  end
+
+
+  get '/tweets/:id/edit' do 
+    if logged_in?
```
Lukeghenco: good use of using your helper method here 👍

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,145 @@ class ApplicationController < Sinatra::Base
...
+      @tweet = Tweet.find(params[:id])
+      if @tweet.user == current_user
```
Lukeghenco: for line 97 I would be more specific. Focus on just the user.id 

```ruby 
if @tweet.user_id == current_user.id
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,145 @@ class ApplicationController < Sinatra::Base
...
+      @tweet = Tweet.find(params[:id])
+      erb :'/tweets/single_tweet'
+    else
+      redirect to "/login"
+    end
+
+  end
+
+
+  post '/tweets/:id' do
```
Lukeghenco: this should be a patch route.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,145 @@ class ApplicationController < Sinatra::Base
...
+    if params[:content] != ""    
+      @tweet.update(content: params[:content])
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    else
+      redirect to "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  post '/tweets/:id/delete' do
```
Lukeghenco: this should be a `delete` route. 

```ruby 
delete '/tweets/:id/delete' do 
```

app/views/tweets/edit_tweet.erb
```diff
@@ -0,0 +1,9 @@
+Edit Tweet:
+
+<form action="/tweets/<%= @tweet.id %>" method="POST">
+
```
Lukeghenco: you will need to add a hidden tag for the patch method to work here

```html
<input id="hidden" type="hidden" name="_method" value="PATCH">
```

app/views/tweets/show_tweet.erb
```diff
@@ -0,0 +1,9 @@
+<h1><%= @user.username %>'s tweets</h1>
```
Lukeghenco: use current_user here.

app/views/tweets/show_tweet.erb
```diff
@@ -0,0 +1,9 @@
...
+
+<ul>
+  <% @user.tweets.each do |tweet| %>
```
Lukeghenco: and current_user here too.

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,7 @@
+<h1>Welcome, <%= @user.username %></h1>
```
Lukeghenco: current_user here as well.

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,7 @@
...
+
+<ul>
+  <% Tweet.all.each do |tweet| %>
```
Lukeghenco: I would change this to be `@tweets` and in the `tweets#index` part of the tweets_controller set the `@tweets = Tweet.all`

Lukeghenco: That will keep the logic and active Record query out of the view.

PR #198, opened 2016-09-02
Project submitted

KerryAlsace:


app/controllers/users_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  get '/signup' do
+    if is_logged_in?
+      redirect to '/tweets'
+    else
+      erb :'/users/create_user'
+    end
+  end
+
+  post '/signup' do
+    @user = User.new
```
Lukeghenco: lets use remove the instance variable here and use User.create here and simplify this
```
user = User.create(username: params[:username], email: params[:email], password: params[:password])
if user.save
  login(user.username, user.password)
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+  end
+
+  ######## DELETE TWEET #########
+
+  post '/tweets/:id/delete' do
+    @tweet = Tweet.find(params["id"])
+    if session["username"] == nil
+      redirect '/login'
+    elsif @tweet.user.username == session["username"]
```
Lukeghenco: i would have this check `tweet.user.id == session[:user_id]` since we are checking for uniqueness of username in our Users model. Right now we could many users with the same username

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    else
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  ######## DELETE TWEET #########
+
+  post '/tweets/:id/delete' do
+    @tweet = Tweet.find(params["id"])
+    if session["username"] == nil
```
Lukeghenco: if is_logged_in? would be a good method here.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      erb :'/tweets/show_tweet'
+    else
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  ######## DELETE TWEET #########
+
+  post '/tweets/:id/delete' do
+    @tweet = Tweet.find(params["id"])
```
Lukeghenco: again we don't need an instance variable here since we aren't going to be showing @tweet in the view.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+  end
+
+  ######## EDIT TWEET #########
+  get '/tweets/:id/edit' do
+    if !is_logged_in?
+      redirect to '/login'
+    end
+    @tweet = Tweet.find(params["id"])
+    if @tweet.user.username == session["username"]
```
Lukeghenco: check for `@tweet.user.id == current_user.id`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect '/tweets'
+    else
+      redirect '/tweets/new'
+    end
+  end
+
+  ######## SHOW TWEET #########
+  get '/tweets/:id' do
+    @tweet = Tweet.find(params["id"])
+    if session["username"] == nil
```
Lukeghenco: good time to use `is_logged_in?` here

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect to '/login'
+    else
+      erb :'/tweets/create_tweet'
+    end
+  end
+
+  post '/tweets' do
+    if !(params["content"] == "")
+      @tweet = Tweet.create(content: params["content"])
+      @tweet.user = User.find_by(username: session["username"])
```
Lukeghenco: search by session[:id] instead here.

PR #199, opened 2016-09-03
Done

tddycks:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
+class TweetsController < ApplicationController
+
+use Rack::Flash
+
+get "/tweets" do 
+  if logged_in?
+    @user = current_user
```
Lukeghenco: you should have access to current_user in the view

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    erb :'/tweets/tweets'
+  else
+    redirect "/login"
+  end
+end
+
+
+get '/tweets/new' do
+  if logged_in?
+    @user = current_user
```
Lukeghenco: you should have access to current_user in the view

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    erb :'/tweets/create_tweet'
+  else
+    redirect "/login"
+  end
+end
+
+post '/tweets' do
+  if !params[:content].empty?
+  current_user.tweets.create(content: params[:content], user_id: session[:user_id])
```
Lukeghenco: great use of the associative create method command, but with this method you don't need to set user_id. This will automatically be done for you.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+  else
+    flash[:message] = "Sorry, you can only edit your tweets."
+    redirect '/login'
+  end
+end
+
+patch '/tweets/:id' do
+  if !params[:content].empty?
+    @tweet = Tweet.find(params[:id])
+    @tweet.content = params[:content]
```
Lukeghenco: you can use the update method here instead 

```
@tweet.update(content: params[:content]
``` 

this will run the save command for you.

app/models/user.rb
```diff
@@ -0,0 +1,17 @@
+class User < ActiveRecord::Base
+
+  has_secure_password
+  has_many :tweets
+  validates_presence_of :email
+  validates_presence_of :username
+  validates_presence_of :password
+
```
Lukeghenco: I would validate uniqueness of username as well

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  if logged_in?
+    redirect '/tweets'
+  else
+  erb :'users/login'
+  end
+end
+
+post "/login" do
+  #raise params.inspect
+  user = User.find_by(username: params[:username])
```
Lukeghenco: what if two users exist with the same username? will this still work?

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,11 @@
+<% if flash.has?(:message) %>
+  <%= flash[:message] %>
+<% end %>
+
+<h1>Welcome, <%= @user.username %></h1>
```
Lukeghenco: current_user.username ?

app/controllers/tweets_controller.rb
tddycks: great thanks.

app/controllers/tweets_controller.rb
tddycks: thanks

app/controllers/tweets_controller.rb
tddycks: Ah cool, good to know. Gotta love ActiveRecord.

tddycks: Definitely makes nicer code. I can then get rid of the argument for the create method:  content: params[:content] and use mass assignment instead:
current_user.tweets.create(params)
thanks.

app/controllers/users_controller.rb
tddycks: That's a great point. I should validate the uniqueness of the username in the user model. So I just added:
validates_uniqueness_of :username
thanks.

app/models/user.rb
tddycks: done

app/views/tweets/tweets.erb
tddycks: That works thanks.

PR #200, opened 2016-09-03
Done.

becky000:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,28 @@
...
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "secret"
+    use Rack::Flash
   end
 
-end
\ No newline at end of file
+  get '/' do
+    @title = "Fwitter - Welcome"
```
Lukeghenco: This is something that I would just hardcode in the view. Instance variable should be used for data stored in the session or pulled from the database that you want to show in the view.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
+class TweetsController < ApplicationController
+
+  get "/tweets" do
+    if logged_in?
+      @title = "Fwitter"
```
Lukeghenco: again not needed this should be hardcoded in the view.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      @tweets = Tweet.all
+      @user = current_user
```
Lukeghenco: you have access to current_user in the view

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      erb :"tweets/index"
+    else
+      flash[:message] = "You must be logged in to view tweets."
+     redirect to "/login"
+    end
+  end
+
+  get "/tweets/new" do
+    if logged_in?
+      @title = "Create Tweet"
```
Lukeghenco: lines 17-18 are not needed again.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      redirect to "/login"
+    end
+  end
+
+  post "/tweets" do
+    if params[:content] == ""
+      redirect to "/tweets/new"
+    else
+      @user = current_user
+      @tweet = Tweet.create(content: params[:content], user_id: @user.id)
```
Lukeghenco: since this is a has_many relationship
refactor lines 30-31 as
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      erb :"tweets/show"
+    else
+      redirect to "/login"
+    end
+  end
+
+  get "/tweets/:id/edit" do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
Lukeghenco: current_user.id

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      redirect "/login"
+    end
+  end
+
+  patch "/tweets/:id" do
+    if params[:content] == ""
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params[:content]
```
Lukeghenco: ```@tweet.update(content: params[:content])```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete "/tweets/:id/delete" do
+    tweet = Tweet.find_by_id(params[:id])
+    if session[:user_id]
```
Lukeghenco: logged_in?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      tweet = Tweet.find_by_id(params[:id])
+      if tweet.user_id == session[:user_id]
```
Lukeghenco: if tweet.user_id == current_user.id

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+      redirect to "/tweets"
+    end
+  end
+
+  post "/signup" do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
+      flash[:message] = "All fields are required."
+      redirect to "/signup"
+    else
+      user = User.new(username: params[:username], email: params[:email], password: params[:password])
```
Lukeghenco: use create here instead of new

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+      if user.save
+        session[:id] = user.id
```
Lukeghenco: your current_user method is using session[:user_id] these should all be the same

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+        redirect to "/tweets"
+      else
+        flash[:message] = "Something went wrong. Please try again."
+        redirect to "/signup"
+      end
+    end
+  end
+
+  post "/login" do
+    @user = User.find_by(username: params[:username])
```
Lukeghenco: is username required to be unique? this could cause a problem if you are allowed to create two unique users with the same username.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+      redirect to "/tweets"
+    else
+      flash[:message] = "Wrong username or password. Please try again."
+      redirect to "/login"
+    end
+  end
+
+  get "/logout" do
+    if logged_in?
+      @title = "Fwitter"
```
Lukeghenco: again, not needed

app/models/user.rb
```diff
@@ -0,0 +1,12 @@
+class User < ActiveRecord::Base
+  has_many :tweets
+  has_secure_password
+
```
Lukeghenco: add Active Record validation for unique password here.

app/views/tweets/index.erb
```diff
@@ -0,0 +1,11 @@
+  <h3>Welcome, <%= @user.username %>!</h3>
```
Lukeghenco: <%= current_user.username %>

app/controllers/application_controller.rb
becky000: Using the variable for title allows me to customize the tab for the current page a user is visiting.

Lukeghenco: this is not a design pattern I would use. The goal is to keep controllers thin.

app/controllers/users_controller.rb
becky000: Added validations in user model for unique email & username, also password confirmation

app/controllers/application_controller.rb
becky000: Thanks for the feedback, Luke.

PR #201, opened 2016-09-06
Done

ThisIsTyDell:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,68 @@
+class TweetController < ApplicationController
+
+  get '/tweets' do
+    if logged_in?
+      @tweets = Tweet.all
+      @users = User.all
```
Lukeghenco: I am not sure why we would want all of the users here?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    if logged_in?
+      erb :'tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] != ""
+      @tweet = Tweet.create(content: params[:content], user_id: current_user.id)
```
Lukeghenco: create a tweet using the associative Active Record methods 
```
```

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  get '/login' do
+    if !logged_in?
+      erb :'users/login'
+    else
+      redirect to '/tweets'
+    end
+  end
+
+  post '/login' do
+    user = User.find_by(username: params[:username])
```
Lukeghenco: does this work if two users have the same username?

app/models/user.rb
```diff
@@ -0,0 +1,7 @@
+class User < ActiveRecord::Base
+  has_many :tweets
+  has_secure_password
+
```
Lukeghenco: add validation here to make sure the a username is unique

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,22 @@
+<h1 class="milk-text">Welcome, <a href="/users/<%= current_user.username %>">@<%= current_user.username %></a>!</h1>
+<section class="first col-1 shadow">
+<div id="details"><h3>Popular Tweeters</h3>
+  <% @users.each do |user| %>
```
Lukeghenco: this seems uneeded.

app/controllers/tweet_controller.rb
ThisIsTyDell: I wanted all of the users available here, so that I can display a list of "Popular Users" on the all tweets page, in addition to displaying a list of tweets. So on the left 1/3 of the browser it shows a list of users. On the right 2/3 of the browser it list all tweets.

Lukeghenco: Great :)

app/controllers/tweet_controller.rb
ThisIsTyDell: Resolved and updated using
`@tweet = current_user.tweets.create(content: params[:content])`

app/models/user.rb
ThisIsTyDell: Resolved - Added validation to only allow unique username and email address to create new User and persist to database. However it seems as if I should also let the user know what happened, like show an error message to them, informing that "name is taken try again"

Lukeghenco: you have access to a @user.errors if you used the Active Record validation

PR #202, opened 2016-09-07
Project submitted

TylerTaylor:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,31 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "password_security"
+  end
+
+  get '/' do
+    if logged_in?
+      @user = current_user
```
Lukeghenco: current_user is already available to us in the views. This is an unneeded instance variable.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,31 @@ class ApplicationController < Sinatra::Base
...
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
+    end
+    
+    def find_tweet(params)
```
Lukeghenco: this would be better placed in the tweets controller, but it really only does one things that is already abstracted with an ActiveRecord method. I would probably not use this.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
+class TweetsController < ApplicationController
+
+  #### TWEETS INDEX ACTION ####
+
+  get '/tweets' do
+    if logged_in?
+      @tweets = Tweet.all
+      @user = User.find(session[:user_id])
```
Lukeghenco: again current_user already does this for us.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do 
+    if params[:content] == ""
+      redirect '/tweets/new'
+    else
+      user = current_user
+      @tweet = Tweet.create(content: params[:content], user_id: user.id)
```
Lukeghenco: 1. @tweet does not need to be an instance variable 
2. we already have current_user so no need for user variable 
3. We should build tweets using the associative create method. 
```ruby 
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      redirect "/tweets/#{@tweet.id}"
```
Lukeghenco: #{tweet.id} now

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+  end
+
+  #### SHOW/READ ACTION ####
+
+  get '/tweets/:id' do
+    if logged_in?
+      @tweet = find_tweet(params)
```
Lukeghenco: just used Tweet.find(params[:id])

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    if params[:content] == ""
+      redirect "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = find_tweet(params)
```
Lukeghenco: again just use `tweet =  Tweet.find(params[:id])`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+      @tweet.content = params[:content]
```
Lukeghenco: lets use the update method in ActiveRecord here so we don't need to use the save method 
```tweet.update(content: params[:content])```

File: app/controllers/tweets_controller.rb [hunk 9]
```diff
@@ -0,0 +1,81 @@
...
+      @tweet.save
+      redirect "/tweets/#{@tweet.id}"
+    end
+  end
+
+  #### DELETE ACTION ####
+
+  delete '/tweets/:id/delete' do
+    @tweet = find_tweet(params)
```
Lukeghenco: again use `tweet =  Tweet.find(params[:id])`

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    else
+      redirect '/tweets'
+    end
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
+      redirect '/signup'
+    else
+      @user = User.create(params)
```
Lukeghenco: user does not need to be an instance variable

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  get '/login' do
+    if !session[:user_id]
+      erb :'/users/login'
+    else
+      redirect '/tweets'
+    end
+  end
+
+  post '/login' do
+    user = User.find_by(username: params[:username])
```
Lukeghenco: does this work if there are two users that have the same username? Have you set up validations in your user model to protect against duplicate usernames?

PR #204, opened 2016-09-13
Project submitted

WilliamJones62:


app/helpers/helpers.rb
```diff
@@ -0,0 +1,15 @@
+class Helpers
+
+  def self.current_user(session_info)
```
Lukeghenco: what we want is for these to be instance methods without arguments.

an example would be 

```ruby 
def current_user
    User.find(session[:id])
end
```  
and 
``` ruby 
def logged_in?
  !!session[:user_id]
end 
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
+class TweetsController < ApplicationController
+
+  get '/tweets' do
+    if session[:user_id]
```
Lukeghenco: this should grab our helper methods 
example 
`if logged_in?`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      @tweets = Tweet.all
+      erb :'tweets/tweets'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/new' do
+    if session[:user_id]
```
Lukeghenco: `if logged_in?`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] == ""
+      redirect to "/tweets/new"
+    else
+      user = User.find_by_id(session[:user_id])
```
Lukeghenco: for line 24 we have access to current_user helper method that does this for us.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      @tweet = Tweet.create(:content => params[:content], :user_id => user.id)
```
Lukeghenco: we should use the associative create method here this would condense lines 24-25 to be this. 
`current_user.tweets.create(content: params[:content])`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  get '/tweets/:id' do
+    if session[:user_id]
```
Lukeghenco: `if logged_in?`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+      erb :'tweets/show'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if session[:user_id]
```
Lukeghenco: if `logged_in?`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
Lukeghenco: `if @tweet.user_id == current_user.id`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    if params[:content] == ""
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params[:content]
```
Lukeghenco: lets use our ActiveRecord update method here 
```
```
this updates and saves in one method.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:user_id]
```
Lukeghenco: if logged_in?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
Lukeghenco: ```if @tweet.user_id == current_user.id```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,57 @@
+class UsersController < ApplicationController
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    erb :'users/show'
+  end
+
+
+  get '/signup' do
+    if !session[:user_id]
```
Lukeghenco: `if !logged_in?`

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    else
+      redirect to '/tweets'
+    end
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
+      redirect to '/signup'
+    else
+      @user = User.new(:username => params[:username], :email => params[:email], :password => params[:password])
```
Lukeghenco: why not just use `User.create()` ?

PR #205, opened 2016-09-17
Project submitted

marcelgalang:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,23 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+
+  helpers do
+		def logged_in?
+			!!session[:user_id]
+		end
+
+		def current_user
+			User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,23 @@ class ApplicationController < Sinatra::Base
...
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
+
+  helpers do
+		def logged_in?
+			!!session[:user_id]
```
Lukeghenco: We could also use our `current_user` method here.

It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+  get '/tweets/new' do
+    if logged_in?
+      erb :'tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post'/tweets' do
+    if params[:content] == ""
```
Lukeghenco: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :content`) to perform these types of checks more easily.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect to '/tweets/new'
+    else
+      user = User.find_by_id(session[:user_id])
```
Lukeghenco: we do have access to `current_user` here. That might be more useful.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      @tweet = Tweet.create(:content=> params[:content], :user_id=> user.id)
```
Lukeghenco: Another way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      erb :'tweets/show_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if session[:user_id]
```
Lukeghenco: we should use our `logged_in?` method here

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
Lukeghenco: `if @tweet.user == current_user` would be more appropriate here.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    if params[:content] == ""
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params[:content]
```
Lukeghenco: a better way to do lines 58-60 would be to use the active record update method and if successful do seperate redirects. 

```ruby
if @tweet.update(content: params[:content])
  redirect to "/tweets/#{@tweet.id}"
else 
  redirect to "/tweets/#{@tweet.id}/edit"
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+   delete '/tweets/:id/delete' do
+     @tweet = Tweet.find_by_id(params[:id])
+
+     if logged_in?
+       if @tweet.user_id == session[:user_id]
```
Lukeghenco: look at comment on line 43

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  get '/signup' do
+    if !logged_in?
+      erb :'users/create_user'
+    else
+      redirect to '/tweets'
+    end
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      redirect to '/signup'
+    else
+      @user = User.new(:username => params[:username], :email => params[:email], :password => params[:password])
```
Lukeghenco: we could just user `User.create` here and initialize and add to the database in one method.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      session[:user_id] = user.id
+
+      redirect  '/tweets'
+    else
+      redirect to '/signup'
+    end
+  end
+
+  get '/logout' do
+    if session[:user_id] != nil
```
Lukeghenco: if `logged_in?` would work here as well.

app/views/users/show.erb
```diff
@@ -0,0 +1,6 @@
+<p> Here are your tweets, <%= @user.username %>
```
Lukeghenco: we should have access to `current_user` here and for line 2 as well

PR #206, opened 2016-09-18
Done

Traci7822:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,77 @@
...
     set :public_folder, 'public'
     set :views, 'app/views'
+    set :session_secret, "password_security"
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  get '/signup' do
```
Lukeghenco: I would move line 16-66 into a UsersController

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,77 @@
...
+    redirect "/login"
+  end
+
+  get '/users/:slug' do
+    user = User.find_by_slug("#{slug}")
+    redirect "/tweets/<%= user.slug %>"
+  end
+
+  helpers do
+    def current_user(session)
```
Lukeghenco: session is already available to this method so on line 70 we can just use this
```ruby 
def current_user 
  User.find(session[:user_id])
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      @user = User.find_by_id(session[:user_id])
+      erb :"/tweets/index"
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/new' do
+    if logged_in?(session)
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: we can use current_user method here.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+    else
+      redirect "/login"
+    end
+  end
+
+  patch '/tweets/:id/edit' do
+    if params[:content] != nil && params[:content] != ""
+      tweet = Tweet.find_by(:id => params[:id])
+      tweet.update(:content => params[:content])
```
Lukeghenco:
```ruby
if tweet.update(content: params[:content])
  redirect to '/tweets'
else 
  redirect to 'tweet/#{tweet.id}/edit'
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
+class TweetController < ApplicationController
+
+  get '/tweets' do
+    if logged_in?(session)
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: Couldn't we just use `current_user` here?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
+class TweetController < ApplicationController
+
+  get '/tweets' do
```
Lukeghenco: For this to be an MVC framework app, we want to load up our data in the controller. 

Do you want to find all the tweets `@tweets = Tweet.all`? Or just the tweets for the user that's logged in `@tweets= current_user.tweets`?

app/controllers/application_controller.rb
Lukeghenco: We also want to make sure to memoize the current_user, that way we don't keep making database calls
`@current_user ||= find_user`

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      erb :'tweets/new'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    if params[:content] != nil && params[:content] != ""
+      user = User.find_by(:id => session[:user_id])
+      tweet = Tweet.create(:content => params[:content], :user_id => user.id)
```
Lukeghenco: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```

or alternatively

```
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      @tweet = Tweet.find_by(:id => params[:id])
+      erb :"tweets/show"
+    else
+      redirect "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?(session)
+      if params[:id].to_i == session[:user_id]
```
Lukeghenco: This is a little odd and i don't think it accomplishes what you intended it to.

Here, `params[:id]` is the tweet id. You are saying that a user can only edit a tweet, it if has the same ID as the user. We really want to be finding the tweet, and then checking if it belongs to the user before letting them edit:

```ruby
@tweet = Tweest.find_by(id: params[:id])
if  @tweet.user == current_user
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      else
+        redirect "/tweets"
+      end
+    else
+      redirect "/login"
+    end
+  end
+
+  patch '/tweets/:id/edit' do
+    if params[:content] != nil && params[:content] != ""
```
Lukeghenco: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :content`) to perform these types of checks more easily.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+        erb :"tweets/edit"
+      else
+        redirect "/tweets"
+      end
+    else
+      redirect "/login"
+    end
+  end
+
+  patch '/tweets/:id/edit' do
```
Lukeghenco: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      tweet = Tweet.find_by(:id => params[:id])
+      tweet.update(:content => params[:content])
+      redirect "/tweets"
+    end
+  end
+
+  delete '/tweets/:id' do
+    if logged_in?(session)
+      tweet = Tweet.find_by(:id => params[:id])
+      if params[:id].to_i == session[:user_id]
```
Lukeghenco: See note on line 42

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+        tweet = Tweet.delete(params[:id])
```
Lukeghenco: We can just do `tweet.delete` here. We already set the variable `tweet` above. Furthermore, you do not go on to use the `tweet` variable set here...

app/models/concerns/validate.rb
```diff
@@ -0,0 +1,9 @@
+module Validate
+  module InstanceMethods
+    def user_params_blank?(params)
```
Lukeghenco: Could you use this in your tweet controller when you edit the tweet?

app/views/tweets/index.erb
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<html>
+    <head>
+      <link rel="stylesheet" type="text/css" href="/style.css">
+    </head>
+    <body>
+    <div class="body">
+      <div class="header">
+        <h1>Welcome, <%= @user.username %></h1>
```
Lukeghenco: You can just use `current_user` here.

app/views/tweets/index.erb
```diff
@@ -0,0 +1,31 @@
...
+      </div>
+      <div class="banner">
+        <h2>Tweets</h2>
+      </div>
+      <div class ="tweet_background">
+        <div class="tweets">
+          <% Tweet.all.each do |tweet| %>
```
Lukeghenco: It's not really within the MVC framework to make a database call (`Tweet.all` is a database call, try it in rails console and you'll see the SQL query!) within the view. We should be loading up this data in the controller.

config.ru
```diff
@@ -5,4 +5,6 @@ if ActiveRecord::Migrator.needs_migration?
 end
 
 use Rack::MethodOverride
-run ApplicationController
\ No newline at end of file
+use UserController
```
Lukeghenco: Where is your UsersController?

config/environment.rb
```diff
@@ -1,6 +1,8 @@
 ENV['SINATRA_ENV'] ||= "development"
 
 require 'bundler/setup'
+require 'bcrypt'
```
Lukeghenco: Do you need `bcrypt` here if it's already in your Gemfile

db/migrate/20160915085131_create_users_table.rb
```diff
@@ -0,0 +1,10 @@
+class CreateUsersTable < ActiveRecord::Migration
+  def change
+    create_table :users do |t|
+      t.string :username
+      t.string :email
+      t.string :password_digest
+      t.integer :tweet_ids
```
Lukeghenco: If you look in your database, what is stored in `tweet_ids` column? What is it used for? 
In a has_many, belongs_to relationship, which model has the foreign key?

PR #207, opened 2016-09-19
Project submitted

bennorris:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "evelynwaugh"
   end
 
-end
\ No newline at end of file
+  helpers do
```
Lukeghenco: you should add a current_user method in your helpers too
your helper methods should look like this. 

```ruby 
def logged_in?
  !!current_user 
end 

def current_user 
  @current_user ||= User.find_by_id(session[:user_id])
end 
```
This will give you access to current_user in the views and your other controllers

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+    end
+
+  end
+
+  get '/' do
+    erb :'home/index'
+  end
+
+  get '/login' do
+    if logged_in?
```
Lukeghenco: good use of your helper methods.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+      end
+    end
+
+  end
+
+  get '/' do
+    erb :'home/index'
+  end
+
+  get '/login' do
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+
+  get '/signup' do
+    if logged_in?
+      redirect to '/tweets'
+    end
+    erb :'home/signup'
+  end
+
+  post '/signup' do
+    if params[:user][:username].empty? || params[:user][:password].empty? || params[:user][:email].empty?
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+      redirect to '/signup'
+    elsif !params[:user][:email].include?("@") || !params[:user][:email].include?(".")
+      redirect to '/signup'
+    else
+      @user = User.create(params[:user])
```
Lukeghenco: we don't need to set an instance variable for a post route that we won't use on the view.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+      session[:user_id] = @user.id
+      redirect to '/tweets'
+    end
+
+  end
+
+  post '/login' do
+
+    @user = User.find_by(username: params[:user][:username])
```
Lukeghenco: see comment on line 48

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+      session[:user_id] = @user.id
+      redirect '/tweets'
+    else
+      redirect '/login'
+    end
+
+  end
+
+  get '/success' do
+    @user = User.find_by_id(session[:user_id])
```
Lukeghenco: we can eliminate this line and just use `<%= current_user %>` in the view

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,84 @@
+require 'pry'
+
+class TweetController < ApplicationController
+
+  get '/tweets' do
+
+    if !logged_in?
+      redirect to '/login'
+    end
+    @user = User.find_by_id(session[:user_id])
```
Lukeghenco: use our `current_user` method.

Lukeghenco: Why aren't we loading our tweets into an instance variable here.

example: 

```
@tweets = Tweet.all
```
we should be separating our logic using the MVC design patter. We should not be making database calls in the view.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if (params[:tweet][:content]).empty?
+      redirect to '/tweets/new'
+    else
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: remember to use current_user method.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      @tweet = Tweet.create(params[:tweet])
+      @user.tweets << @tweet
```
Lukeghenco: for Lines 27-28 we should be using the associative build method in Active Record. 

```ruby 
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      redirect to '/tweets'
+    end
+  end
+
+  get '/tweets/:id' do
+    if logged_in?
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: not needed with current_user method.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+          erb :'tweets/show_individual_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
```
Lukeghenco: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      erb :'tweets/edit'
+    else
+      redirect to '/login'
+    end
+  end
+
+
+
+  post '/tweets/:id' do
+    @tweet = Tweet.find_by_id(params[:id])
```
Lukeghenco: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  get '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
+  end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: not needed

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      if @tweet.user_id == @user.id
```
Lukeghenco: `if @tweet.user == current_user`

app/views/tweets/show.erb
```diff
@@ -0,0 +1,20 @@
+<h1>Welcome, <%=@user.username%></h1>
+
+<h3>Your Fwitter digest:</h3>
+
+<%Tweet.all.reverse.each do |tweet|%>
```
Lukeghenco: this logic should be delegated to the COntroller

app/views/tweets/show.erb
```diff
@@ -0,0 +1,20 @@
...
+  <%user = User.find_by_id(tweet.user_id)%>
```
Lukeghenco: see note on line 5

app/views/tweets/show.erb
```diff
@@ -0,0 +1,20 @@
...
+  <a href="/users/<%=user.slug%>"><%=user.username%></a>: <%=tweet.content%> <br>
+  <%end%>
+
+<h3>Edit your Tweets:</a></h3>
+<%x=1%>
+<%@user.tweets.each do |tweet|%>
```
Lukeghenco: this should be save to another instance as well
```
@user_tweets = current_user.tweets
```
as we are still making database calls

PR #209, opened 2016-09-20
Done

sandyvern:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,23 @@ class ApplicationController < Sinatra::Base
...
+    erb :index
+  end
+
+  helpers do
+
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
  @current_user ||= User.find_by_id(session[:id) 
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,23 @@ class ApplicationController < Sinatra::Base
...
 
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+    if logged_in?
+      @tweets = Tweet.all
+      erb :'tweets/tweets'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/new' do
+    if logged_in?
```
Lukeghenco: good use of our helper methods

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] == ""
+      redirect to "/tweets/new"
+    else
+      @tweet = current_user.tweets.create(content: params[:content])
```
Lukeghenco: great use of an associative build method.

Lukeghenco: not necessary to make this an instance method since we won't be loading it up on the view.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+       erb :'tweets/edit_tweet'
+      else
+        redirect to '/tweets'
+      end
+    else
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+    if params[:content] == ""
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params[:content]
```
Lukeghenco: I would use `update` here instead. 
```ruby
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
```
Lukeghenco: we have no need of an instance variable since this will not be use in the view.

Lukeghenco: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+ get '/signup' do
+   if !logged_in?
+     erb :'users/create_user', locals: {message: "Please sign up before you sign in"}
+   else
+     redirect to '/tweets'
+   end
+ end
+
+ post '/signup' do
+   if params[:username] == "" || params[:email] == "" || params[:password] == ""
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+     redirect to '/signup'
+   else
+     @user = User.create(:username => params[:username], :email => params[:email], :password => params[:password])
```
Lukeghenco: no need to make @user an instance variable.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+ get '/login' do
+   if !logged_in?
+     erb :'users/login'
+   else
+     redirect '/tweets'
+   end
+ end
+
+ post '/login' do
+   user = User.find_by(:username => params[:username])
```
Lukeghenco: what if two users have the same username would this still work correctly? Look into Active Record validations for uniqueness

PR #210, opened 2016-09-22
Project submitted

bmolina-nyc:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,44 @@
...
+  end
+
+
+helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find_by_id(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,44 @@
...
+  # this is an index page 
+  # should expect logic for redirects to login, and signup
+  get '/' do  
+    erb :index
+  end
+
+
+helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+    enable :sessions
+    set :session_secret, "secret"
+  end
+
+  # the tweets index page
+  # this appears to be the users homepage once they're signed in
+  # can grab a users specific tweets - a user has many tweets
+  get '/tweets' do 
+
+      if current_user
```
Lukeghenco: if logged_in?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+        @user = current_user
```
Lukeghenco: we don't need to set an instance variable to current_user. we have access to current_user in the view already.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+        erb :'tweets/tweets'
+      else
+        flash[:message] = "You must be logged in"
+        redirect to '/login'
+      end
+  end
+
+   get '/tweets/new' do 
+    @user = current_user
```
Lukeghenco: see note on line 20

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+    if @user
+      erb :'/tweets/create_tweet'
+    else
+      flash[:message] = "You must login"
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    @user = current_user
```
Lukeghenco: see comment on line 20

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    if params[:content].empty?
+      flash[:message] = "Cannot create an empty tweet!"
+      redirect '/tweets/new'
+    else
+      @tweet = Tweet.new(content: params[:content], user_id: @user.id)
```
Lukeghenco: for lines 46-29

The best way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+  get '/tweets/:id/edit' do 
+
+    if !logged_in?
+      flash[:message] = "You gotta login to edit tweets!"
+      redirect to '/login' 
+    end
+
+    @tweet = Tweet.find(params[:id])
+    
+    if @tweet.user_id == current_user.id
```
Lukeghenco: or `if @tweet.user == current_user`

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    if logged_in?  # if a user session isnt already live, populate the signup
+      redirect to '/tweets'
+    else
+      erb :'/signup'
+    end
+  end
+
+  # catches the signup params - creates a new user - and logs them into the tweets homepage
+  post '/signup' do 
+    if params.values.any? { |el| el.empty? }
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      flash[:message] = "You need to enter all fields to signup!"
+      redirect to '/signup'
+    else 
+      @user = User.create(username: params["username"], email: params["email"], password: params["password"])
```
Lukeghenco: create automatically saves

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,24 @@
...
+
+<h4><a href="/users/<%=@user.slug%>">Your Homepage</a></h4>
+
+<h4><a href="/tweets/new">Make a Tweet</a></h4>
+
+
+<br>
+
+<h2>All Tweets - Click to go to edit page</h2>
+  <ul><% Tweet.all.each do |tweet| %>
```
Lukeghenco: In the MVC design pattern we should avoid database calls in the view. You would be better served setting this to an instance variable in the `get '/tweets' do` method

```ruby
@tweets = Tweet.all
```

app/views/users/user_homepage.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Welcome, <%= @user.username %> - These are your tweets!</h1>
+
+  <% if !@user.tweets.empty? %>
+    <% @user.tweets.each do |tweet| %>
```
Lukeghenco: In the MVC design pattern we should avoid database calls in the view. You would be better served setting this to an instance variable in the get `'/users/:id'` do method
```ruby
@tweets = current_user.tweets
```

bmolina-nyc:
Thanks for this note Luke - I'll get to this at some point over the next few days and update you when I'm all set.

Regards,

Bruce

PR #211, opened 2016-09-25
Done

ms94ap:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "evelynwaugh"
   end
 
+  helpers do
+    def logged_in?
+      if session[:user_id]
```

```ruby
def current_user
end
```

It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+
+  get '/signup' do
+    if logged_in?
+      redirect to '/tweets'
+    end
+    erb :'home/signup'
+  end
+
+  post '/signup' do
+    if params[:user][:username].empty? || params[:user][:password].empty? || params[:user][:email].empty?
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+      session[:user_id] = @user.id
+      redirect '/tweets'
+    else
+      redirect '/login'
+    end
+
+  end
+
+  get '/success' do
+    @user = User.find_by_id(session[:user_id])
```
Lukeghenco: we can use our `current_user` helper method here

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,79 @@
...
+      end
+    end
+
+  end
+
+  get '/' do
+    erb :'home/index'
+  end
+
+  get '/login' do
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
+
+class TweetController < ApplicationController
+
+  get '/tweets' do
+
+    if !logged_in?
+      redirect to '/login'
+    end
+    @user = User.find_by_id(session[:user_id])
```
Lukeghenco: see note on line 69

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if (params[:tweet][:content]).empty?
+      redirect to '/tweets/new'
+    else
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: see note on line 69

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      @tweet = Tweet.create(params[:tweet])
```
Lukeghenco: The best way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      @user.tweets << @tweet
+      redirect to '/tweets'
+    end
+  end
+
+  get '/tweets/:id' do
+    if logged_in?
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: see note on line 69

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  get '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
+  end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: not needed

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      if @tweet.user_id == @user.id
```
Lukeghenco: a better way to do this is

```
if @tweet.user == current_user
```

app/views/home/success.erb
```diff
@@ -0,0 +1,7 @@
+
+<h1>Welcome, <%=@user.username%></h1>
```
Lukeghenco: we can do this here `<%= current_user.username %>`

app/views/tweets/show.erb
```diff
@@ -0,0 +1,21 @@
+
+<h1>Welcome, <%=@user.username%></h1>
+
+<h3>Your Fwitter digest:</h3>
+
+<%Tweet.all.reverse.each do |tweet|%>
```
Lukeghenco: we should leave database logic out of the view and set this to an instance variable in the TweetsController

`@tweets = Tweet.all`

app/views/tweets/show.erb
```diff
@@ -0,0 +1,21 @@
...
+  <%user = User.find_by_id(tweet.user_id)%>
+  <a href="/users/<%=user.slug%>"><%=user.username%></a>: <%=tweet.content%> <br>
+  <%end%>
+
+<h3>Edit your Tweets:</a></h3>
+<%x=1%>
+<%@user.tweets.each do |tweet|%>
```
Lukeghenco: Again, we should leave database logic out of the view and set this to an instance variable in the TweetsController

`@user_tweets = current_user.tweets`

Lukeghenco (commented):
Hi Andrew,

Great job on your Fwitter project. I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Please notify me when you have completed the fixes or if you have any questions by email: luke.ghenco@flatironschool.com or on slack: (@luke).

Regards,

{{ INSTRUCTOR NAME }}

PR #212, opened 2016-10-03
Done

WilliamMena:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "secret"
   end
 
+  get '/' do 
+    erb :index
+  end
+
+  get '/signup' do
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,15 @@
+class Helper < ApplicationController
+
+  def self.current_user(session)
```
Lukeghenco: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,15 @@
...
+    User.find(session[:user_id])
+  end
+
+  def self.logged_in?(session)
```
Lukeghenco: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
+  get '/signup' do 
+    if session[:user_id]
+      redirect to 'tweets'
+    else
+      erb :'/users/create_user'
+    end
+  end
+
+  post '/signup' do
+    if params["username"].empty? || params["password"].empty? || params["email"].empty?
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
+    enable :sessions
+    set :session_secret, "secret"
   end
 
+  get '/' do 
+    erb :index
+  end
+
+  get '/signup' do 
+    if session[:user_id]
```
Lukeghenco: You can use your `logged_in?` helper method here, which is preferable to referring to the session object directly.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
+  get '/login' do 
+    if session[:user_id]
+      redirect to '/tweets'
+    else
+      erb :'/users/login'
+    end
+  end
+
+  post '/login' do
+    user = User.find_by(username: params["username"])
```
Lukeghenco: do you have a validation on the user model, so that when a user is created it has a unique username? Otherwise we might run into a problem here.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
+    if user && user.authenticate(params["password"])
```
Lukeghenco: see note on line 17

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
+      session[:user_id] = user.id
+      redirect to '/tweets'
+    else
+      "Wrong Password"
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets' do
+    if session[:user_id]
```
Lukeghenco: see note on line 17

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: line 55 and 56 are not needed. We are able to just call `current_user` in the view.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  get '/logout' do
+    session.clear
+    redirect to '/login'
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by_id(session["user_id"])
+    @view = User.find_by_slug(params[:slug])
```
Lukeghenco: why are you setting an @view here?

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params["content"].empty?
+      #raise alert here, look back at last lesson to raise alert.
+      redirect to '/tweets/new'
+    else
+      @user = Helper.current_user(session)
```
Lukeghenco: for lines 88 - 90 I would just do this. 

```
current_user.tweets.create(params)
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,138 @@ class ApplicationController < Sinatra::Base
...
+      redirect to "/tweets/#{@tweet.id}/edit"
+    else
+      @tweet.content = params["content"]
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
```
Lukeghenco: we don't need to use an instance variable here as we are not accessing this in the view.

app/helpers/helpers.rb
```diff
@@ -0,0 +1,15 @@
+class Helper < ApplicationController
+
```
Lukeghenco: It would be easier to leave some of these methods in your ApplicationController 

```ruby 
helpers do 
   def logged_in?

  end 

  def current_user 

  end 

end
```

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,29 @@
+<% if Helper.logged_in?(session) %>
+  <h1>Welcome, <%= @user.username.capitalize %></h1>
+
+  <h3>Feed</h3>
+
+  <% User.all.map do |u| %>
```
Lukeghenco: In a MVC pattern we try to avoid database calls in the View. I would set this to `@users = User.all` in the `get '/tweets' do` method

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,29 @@
...
+      <b><%= u.username %></b>: <%= t.content %>
+      <br>
+    <% end %>
+
+  <% end %>
+<% end %> <!-- Ends the logged in content -->
+
+
+  <h1><%= @view.username %></h1>
+<% @view.tweets.each do |t| %>
```
Lukeghenco: you can just call `current_user.tweets` here

PR #213, opened 2016-10-04
Lani

Lani1321:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,22 @@ class ApplicationController < Sinatra::Base
...
+  get '/' do
+    erb :index
+  end
+
+ helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
  @current_user ||= User.find_by_id(session[:user_id])
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do 
+    if params[:content] == ""
+      redirect "/tweets/new"
+    else
+      user = User.find_by_id(session[:user_id])
```
Lukeghenco: you should have access to `current_user` here

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      @tweet = Tweet.create(:content => params[:content], :user_id => user.id)
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      erb :'tweets/show_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
Lukeghenco: It would be better use current_user here `if @tweet.user == current_user`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    if params[:content] == ""
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params[:content]
```
Lukeghenco: you could use `@tweet.update(params)`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```
Lukeghenco: see comment on line 44

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      @tweet.content = params[:content]
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:user_id]
+      @tweet = Tweet.find_by_id(params[:id])
```
Lukeghenco: you already have an instance of @tweet saved on line 66

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,55 @@
+class UsersController < ApplicationController
+
+  get '/signup' do 
+    if !session[:user_id]
```
Lukeghenco: `if !logged_in?`

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      erb :"users/create_user"
+    else
+      redirect "/tweets"
+    end
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    else
+      redirect "/tweets"
+    end
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == "" 
+      redirect "/signup"
+    else
+      @user = User.new(:username => params[:username], :email => params[:email], :password => params[:password])
```
Lukeghenco: we could just pass in params as well. 

Example: 
```ruby
user = User.create(params)
session[:user_id] = user.id
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  get '/login' do
+    if !logged_in?
+       erb :'/users/login'
+    else
+       redirect to '/tweets'
+    end
+  end
+
+  post '/login' do
+    user = User.find_by(:username => params[:username])
```
Lukeghenco: is there a validation for users so that two users can not make an account with the same username? if so this will not work correctly.

Lani1321: I know how to validate this, but I can't seem to get the error message to pop up without using rack flash.  Is there an easier way to implement the error message?

Lukeghenco: you can pass it as a parameter `error?` but rack flash is probably easier

app/controllers/tweets_controller.rb
Lani1321: done

app/controllers/tweets_controller.rb
Lani1321: oops...line 66 is now deleted

app/controllers/tweets_controller.rb
Lani1321: done

app/controllers/users_controller.rb
Lani1321: done

app/controllers/users_controller.rb
Lani1321: done

app/controllers/users_controller.rb
Lani1321: done

Lani1321:
Hi Luke!
We just finished refactoring.  Let us know if you have any other comments.
Thanks for your suggestions!

Lani and Rafeeq

PR #214, opened 2016-10-04
Project submitted

lair001:


app/controllers/application_controller.rb
```diff
@@ -2,9 +2,140 @@
...
+		set :views, 'app/views'
+		enable :sessions
+		set :session_secret, "password_security"
+	end
+
+	get '/' do 
+		erb :index
+	end
+
+	get '/tweets' do
```

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,140 @@
...
+	get '/tweets' do 
+		if logged_in?
+			erb :'tweets/tweets'
+		else
+			redirect '/login'
+		end
+	end
+
+	get '/signup' do 
+		if self.logged_in?
```
Lukeghenco: no reason to call self here

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,140 @@
...
+	end
+
+	get '/users/:slug' do 
+		@user = User.find_by_slug(params[:slug])
+		erb :'users/user_tweets'
+	end
+
+
+	post '/signup' do 
+		@user = User.new
```
Lukeghenco: you could also just pass params as the argument 
example: 
```
user = User.new(params)
``` 
I would refrain from creating an instance variable here as you will not need to access it in the view.

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,140 @@
...
+		@user.username = params[:username]
+		@user.email = params[:email]
+		@user.password = params[:password]
+		if @user.save
+			session[:id] = @user.id
```
Lukeghenco: I would set session as session[:user_id] instead of [:id] as this is a user account.

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,140 @@
...
+			redirect '/tweets'
+		else
+			redirect '/signup'
+		end
+	end
+
+	post '/login' do
+		@user = User.find_by(username: params[:username])
```
Lukeghenco: is their validations in place to make sure that two users can not make an account with the same username? Otherwise this could cause problems

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,140 @@
...
+	get '/tweets/new' do
+		if logged_in?
+			erb :'tweets/create_tweet'
+		else
+			redirect '/login'
+		end
+	end
+
+	post '/tweets' do 
+		if !params[:content].nil? && params[:content] != ""
```
Lukeghenco: we could set up ActiveRecord validations in the Tweet class for this as well.

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,140 @@
...
+			@tweet = Tweet.new
```
Lukeghenco: The best way to do this is to make use of the `has_many` relationship on User:
```
@tweet = current_user.tweets.build(content: params[:content])
```

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,140 @@
...
+	delete '/tweets/:id/delete' do 
+		@tweet = Tweet.find(params[:id])
+		@tweet.delete if @tweet.user == current_user
+		redirect '/tweets'
+	end
+
+	patch '/tweets' do 
+		if !params[:content].nil? && params[:content] != ""
+			@tweet = Tweet.find(params[:id]) 
+			@tweet.content = params[:content]
```
Lukeghenco: `@tweet.update(params)` would be sufficient here.

app/controllers/application_controller.rb
```diff
@@ -2,9 +2,140 @@
...
+			redirect "/tweets/#{params[:id]}/edit"
+		end
+	end
+
+	helpers do
+		def logged_in?
+		  !!session[:id]
+		end
+
+		def current_user
```

```ruby
def current_user
end
```

app/views/users/user_tweets.erb
```diff
@@ -0,0 +1,5 @@
+<h2>These tweets were written by <%=@user.username%>!</h2>
+<% @user.tweets.each do |tweet| %>
```
Lukeghenco: You should avoid database calls on the front end. We should set `@user_tweets = @user.tweets` in the TweetsController

Lukeghenco (commented):
Hi Samuel,

Great job on your Fwitter project. I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Please notify me when you have completed the fixes or if you have any questions by email: luke.ghenco@flatironschool.com or on slack: (@luke).

Regards,

Luke Ghenco

app/controllers/application_controller.rb
lair001: I would tend to agree, but for a lab, the tests often constrain the data structure and field ids.  Making this change would cause some tests to fail.

app/controllers/application_controller.rb
lair001: Apparently, you still need to follow a .update with .save so this would only replace @tweet.content = params[:content] with @tweet.update(params).  The latter would also attempt to change the id.  I'm not sure I can agree with this one.

app/views/users/user_tweets.erb
lair001: done

app/controllers/application_controller.rb
lair001: done

app/controllers/application_controller.rb
lair001: done

app/controllers/application_controller.rb
lair001: done

app/controllers/application_controller.rb
lair001: Except for the instance variable part.  Although it is not needed in a view currently, that could change in future versions, and simply making everything an instance variable in the beginning could save some time and confusion later.  Is this approach possibly a security concern?

app/controllers/application_controller.rb
lair001: I tried two solutions and both break your tests:
The first:
	post '/login' do
		@user = User.find_by(username: params[:username], email: params[:email])
		if @user && @user.authenticate(params[:password])
			session[:id] = @user.id
			redirect '/tweets'
		else
			redirect '/'
		end
	end
The second:

	post '/signup' do 
		@user = User.new(params)
		if User.find_by_slug(@user.slug).nil? && @user.save
			session[:id] = @user.id
			redirect '/tweets'
		else
			redirect '/signup'
		end
	end

app/controllers/application_controller.rb
lair001: done

app/controllers/application_controller.rb
lair001: done

app/controllers/application_controller.rb
lair001: done

lair001:
Dear Luke,

I am done with my reply to your suggestions.  Most are implemented.  A few
I commented on.  See pull request for details.

Sincerely,


Samuel Lair

app/controllers/application_controller.rb
Lukeghenco: So the Active Record `update` method will only `update` the `attributes` that would change, in this situation the `params` that are passed. Update also automatically saves to the database, so it will return with a true or false just as save returns with true or false. The id will not have changed, so it will not be updated.

app/controllers/application_controller.rb
lair001: The thing is, if you want to prevent any problems, you would have to prevent the creation of usernames that resolve to the same slug.  E.g. 'The Great' and 'The-Great' would both resolve to the same slug 'the-great' while passing a validation based on username.  To use ActiveRecord validation, the only solution I can think of is to make the slug a database column, which, of course increases the size of the database.

app/controllers/application_controller.rb
lair001: done

app/controllers/application_controller.rb
lair001: Alright, even if you implement a ActiveRecord validation based on username by adding
validates :username, uniqueness: true
in users.rb, the tests still break in the same way when trying to validate by slug with my code above:
  1) ApplicationController Signup Page does not let a logged in user view the signup page
     Failure/Error: expect(last_response.location).to include('/tweets')
       expected nil to include "/tweets", but it does not respond to `include?`
     # ./spec/controllers/application_controller_spec.rb:71:in `block (3 levels) in <top (required)>'

Lukeghenco: you shouldn't be finding by slug

with you user model now validating uniqueness of :username 

you will know just have this code. 

```

lair001:
Even if you do it the way you are suggesting by placing:
validates :username, uniqueness: true
in user.rb, you still break the tests (in the same way my validation by
slug did):

  1) ApplicationController Signup Page does not let a logged in user view
the signup page
     Failure/Error: expect(last_response.location).to include('/tweets')
       expected nil to include "/tweets", but it does not respond to
`include?`
     # ./spec/controllers/application_controller_spec.rb:71:in `block (3
levels) in <top (required)>'

lair001:
I.e. I understand where you are trying to go, but I don't want to rewrite
the test suite.

PR #216, opened 2016-10-06
Project submitted

nicolefederici:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,67 @@ class ApplicationController < Sinatra::Base
...
     set :views, 'app/views'
+    enable :sessions unless test?
+    set :session_secret, "secret"
   end
 
+  get '/' do
+  erb :index
+  end
+
+  get '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,67 @@ class ApplicationController < Sinatra::Base
...
+    erb :'tweets/create_tweet'
+  end
+
+  post '/tweets' do
+    tweet = Tweet.create(params)
```
Lukeghenco: the `create()` method already commits to the database so `.save` accomplishes nothing here for us.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,67 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions unless test?
+    set :session_secret, "secret"
   end
```
Lukeghenco: We should add some `helper methods here to simplify some of our code below 

```ruby 

helpers do 
  
  def logged_in? 
    !!current_user 
  end 

  def current_user 
    @current_user ||= User.find(session[:user_id]) if session[:user_id]
  end 
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,67 @@ class ApplicationController < Sinatra::Base
...
+    redirect('/tweets')
+  end
+
+  get '/tweets' do
+    @tweets = Tweet.all
+    erb :'tweets/tweets'
+  end
+
+  get '/signup' do
+    if !session
```
Lukeghenco: we can now use our new helper method on line 11

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,67 @@ class ApplicationController < Sinatra::Base
...
+     erb :'users/create_user'
+    else
+    redirect('/tweets')
+    end
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,67 @@ class ApplicationController < Sinatra::Base
...
+  end
+end
+
+
+  get '/login' do
+    erb :'users/login'
+  end
+
+   post '/login' do
+    @user = User.find_by(username: params[:username])
```
Lukeghenco: I would set up a validation in your User model for a username to be unique before it is saved to the database.

PR #217, opened 2016-10-06
Done

bedle88:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+
+  helpers do
+    def logged_in?
+
+      !!session[:id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  get "/logout" do
+    session.clear
+    redirect "/login"
+  end
+
+
+  helpers do
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+  post '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
+    if params[:content].empty?
+      redirect "/tweets/#{params[:id]}/edit"
+    else
+      @tweet.update(content: params[:content])
+      @tweet.save
```
Lukeghenco: the `update` method already commits to the database so we don't need to run save after update

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
+      erb :'/tweets/show_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in? && Tweet.find(params[:id]) != nil
```
Lukeghenco: just check for `if logged_in?` don't make two database calls for Tweet

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+    else
+      @tweet = Tweet.create(content: params[:content], user_id: session[:id])
+      @tweet.save
+    end
+    redirect "/tweets"
+  end
+
+  get '/tweets' do
+    if logged_in?
+      @user = User.find(session[:id])
```
Lukeghenco: great time to use our `current_user` method.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+  get '/login' do
+    if logged_in?
+      redirect '/tweets'
+    else
+      erb :'/users/login'
+    end
+  end
+
+  post '/login' do
+    @user = User.find_by(username: params[:username])
```
Lukeghenco: you should have a ActiveRecord validation on your user model so that a username is unique when a user signs up.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,140 @@ class ApplicationController < Sinatra::Base
...
+  get '/signup' do
+    if logged_in?
+      redirect '/tweets'
+    else
+      erb :'/users/create_user'
+    end
+  end
+
+  post '/signup' do
+    if params[:username].empty? || params[:password].empty? || params[:email].empty?
```

app/models/user.rb
```diff
@@ -0,0 +1,21 @@
+class User < ActiveRecord::Base
+has_many :tweets
+
+  def slug
+    self.username.downcase.gsub(" ", "-")
+  end
+
+  def authenticate(password)
```
Lukeghenco: if you add the `has_secure_password`macro method to this class that `bcrypt` supplies it includes the authenticate method for you.

PR #218, opened 2016-10-10
Done

Alicekb:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,24 @@
...
+  end
+
+  get '/' do
+    erb :index
   end
 
-end
\ No newline at end of file
+  helpers do
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,24 @@
...
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      @tweets = Tweet.all
+      erb :'tweets/tweets'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/new' do
+    if logged_in?
+      erb :'tweets/create_tweet'
```
Lukeghenco: we should keep with the design patterns for CRUD with MVC and have set names for these files 

- index.erb
- show.erb
- new.erb
- edit.erb

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    if params[:content] == ""
+      redirect "/tweets/new"
+    else
+      @user = User.find_by_id(session[:user_id])
```
Lukeghenco: you should have access to `current_user` here

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      @tweet = Tweet.new(content: params[:content],user_id: @user.id)
```
Lukeghenco: use the associative Active Record methods. We should also just go ahead and create his instead of calling save on line 41

`current_user.tweets.create(content: params[:content])`

this will automatically associated the user_id with current_user.id

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+  end
+
+  patch '/tweets/:id/edit' do
+    @tweet = Tweet.find_by_id(params[:id])
+    if current_user.id = @tweet.user.id
+      if params[:content] == ""
+        redirect "/tweets/#{@tweet.id}/edit"
+      else
+        @tweet.content = params[:content]
```
Lukeghenco: we can use `@tweet.update(content: params[:content])` here instead of setting the attribute setter method and calling save.

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    erb :'users/show'
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
```
Lukeghenco: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    else
+      @user = User.new(:username => params[:username], :email => params[:email], :password => params[:password])
+      @user.save
+      session[:user_id] = @user.id
+      redirect '/tweets'
+    end
+  end
+
+  post '/login' do
+    user = User.find_by(:username => params[:username])
```
Lukeghenco: I would set up a validation on your user model for the uniqueness of username, so that no two users can have the same username

PR #219, opened 2016-10-12
Done

auranbuckles:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,79 @@ class ApplicationController < Sinatra::Base
...
+    @user = User.find(params[:slug])
+    erb :'/tweets/show'
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,79 @@ class ApplicationController < Sinatra::Base
...
+    redirect "/login"
+  end
+
+  get '/users/:slug' do
+    @user = User.find(params[:slug])
+    erb :'/tweets/show'
+  end
+
+  helpers do
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,79 @@ class ApplicationController < Sinatra::Base
...
+    @user = User.find_by(:username => params[:username])
+    if @user && @user.authenticate(params[:password])
+        session[:user_id] = @user.id
+        redirect "/tweets"
+    else
+        redirect "/failure"
+    end
+  end
+
+  # get "/success" do
```
Lukeghenco: I would remove out old comment out code. Make it look a lot cleaner.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,79 @@ class ApplicationController < Sinatra::Base
...
+  get "/login" do
+    if logged_in?
+      redirect "/tweets"
+    else
+      erb :login
+    end
+  end
+
+  post "/login" do
+    @user = User.find_by(:username => params[:username])
```
Lukeghenco: You should create a validation on user model to verify that when a User is created that the username is unique on the database.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,79 @@ class ApplicationController < Sinatra::Base
...
+  get "/signup" do
+    if logged_in?
+      redirect "/tweets"
+    else
+      erb :signup
+    end
+  end
+
+  post "/signup" do
+    user = User.new(:username => params[:username], :email => params[:email], :password => params[:password])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,79 @@ class ApplicationController < Sinatra::Base
...
+    if user.save && user.username != "" && user.email != "" && user.password != ""
+        redirect "/tweets"
```
Lukeghenco: if a user is saved you should set the session[:user_id]

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,79 @@ class ApplicationController < Sinatra::Base
...
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "password_security"
+  end
+
+  get "/" do
+    erb :index
+  end
+
+  get "/signup" do
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+			@user = User.find(session[:user_id])
+			erb :'/tweets/index'
+		else
+			redirect "/login"
+		end
+	end
+
+	post '/tweets' do
+    if params[:content] != ""
+      @tweet = Tweet.new(content: params[:content])
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+  	if logged_in?
+      @tweet = Tweet.find(params[:id])
+      erb :'/tweets/edit'
+    else
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
```
Lukeghenco: you should do a second security check here to:

1: make sure a user is `logged_in?`
2. make sure that the `@tweet.user == current_user`

PR #220, opened 2016-10-13
Done

jmb521:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,182 @@ class ApplicationController < Sinatra::Base
...
+    erb :index
+  end
+
+  get '/tweets' do
+# <<<<<<< HEAD
+
+
+# =======
+    if is_logged_in?
+      @user = current_user
```
Lukeghenco: We have access to current_user in the view already

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,182 @@ class ApplicationController < Sinatra::Base
...
+      redirect to '/tweets'
+    end
+  end
+
+  post '/signup' do
+
+    if is_logged_in?
+      redirect '/tweets'
+    else
+      if !params[:username].empty? && !params[:email].empty? && !params[:password].empty?
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,182 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+
+
+  patch '/tweets/:id' do
+    if params[:content] == "" || params[:content] == nil
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params[:content]
```
Lukeghenco: `@tweet.update(content: params[:content])` would work here instead of setting the value and then saving

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,182 @@ class ApplicationController < Sinatra::Base
...
+    if is_logged_in?
+      session.destroy
+      redirect to '/login'
+
+    else
+        redirect to '/'
+    end
+  end
+  get '/users/:slug' do
+    @user = current_user.username.slug
```
Lukeghenco: user = User.find_by_slug(slug: params[:slug])

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,182 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+
+  helpers do
+
+    def is_logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,182 @@ class ApplicationController < Sinatra::Base
...
+  get '/users/:slug' do
+    @user = current_user.username.slug
+    @tweets = Tweet.find_by(@user.id)
+    erb :'users/show'
+  end
+
+
+  helpers do
+
+    def is_logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

PR #222, opened 2016-10-15
Done

SnowboardTechie:


app/controllers/application_controller.rb
```diff
@@ -3,8 +3,14 @@
...
     set :public_folder, 'public'
     set :views, 'app/views'
   end
 
-end
\ No newline at end of file
+  get '/' do
+    erb :'application/index'
+  end
+
```
Lukeghenco: so we would want to set up some helper methods here to handle `current_user` and `logged_in?` so that we do not have to keep checking session in the controllers 

```ruby
helpers do 

  def logged_in?
    !!current_user 
  end 

  def current_user 
  end 
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
+class TweetsController < ApplicationController
+  get '/tweets' do
+    if !session[:id].nil?
```
Lukeghenco: good chance to use our new `logged_in?` method, and for any other `session[:id]` checks in your application.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+      @user = User.find_by(session[:id])
```
Lukeghenco: we should use our new `current_user` for this line of code, and for any other `session[:id]` checks in your application.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    if params["tweet"]["content"].nil? || params["tweet"]["content"] == ""
+      redirect '/tweets/new'
+    else
+      user = User.find_by(session[:id])
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+      tweet = Tweet.new(params["tweet"])
+      user.tweets << tweet
+      redirect to("/tweets/#{tweet.id}")
+    end
+  end
+
+  get '/tweets/:id' do
+    user = User.find_by(session[:id])
+    @tweet =  Tweet.find_by(params[:id])
+    if !session[:id].nil? && @tweet.user_id == user.id
```
Lukeghenco: `if logged_in? and @tweet.user == current_user`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+      erb :'tweets/show'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    @tweet = Tweet.find_by(params[:id])
```
Lukeghenco: `.find_by_id` would be the best syntax here.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  get '/signup' do
+    if session[:id].nil?
+      erb :'users/signup'
+    else
+      redirect '/tweets'
+    end
+  end
+
+  post '/signup' do
+    if (params[:username].empty? || params[:password].empty? || params[:email].empty?)
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      redirect '/signup'
+    else
+      user = User.create(params)
```
Lukeghenco: if would wrap this in a build step to verify this saves to the database and offer different redirect depending on that 
```ruby 
user = User.new(params)
if user.save 
  session[:user_id] = user.id 
  redirect '/tweets'
else 
  redirect '/signup'
end 
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      redirect '/tweets'
+    else
+      erb :'users/login'
+    end
+  end
+
+  post '/login' do
+    user = User.find_by(username: params[:username])
+    session[:id] = user.id
+    if user.authenticate(params[:password])
```
Lukeghenco: we should check to make sure a user exists as well as authenticate the password. we should also set session[:user_id] once we authenticate too.

```ruby 
if user && user.authenticate(params[:password]
  session[:user_id] = user.id 
  redirect '/tweets'
```

app/views/tweets/index.erb
```diff
@@ -0,0 +1,6 @@
+Welcome, <%= @user.username %>
+<% Tweet.all.each do |tweet| %>
```
Lukeghenco: With a MVC design patter you want to avoid Database calls in the view. The controller should interact with the model for this query instead 

```ruby 
def index 
  @tweets = Tweet.all 
end 
```

PR #223, opened 2016-10-16
Done

BriOD:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,28 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,28 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      redirect '/tweets'
+    else
+      erb :'index'
+    end
+  end
+
+
+  helpers do
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  get '/tweets/new' do
+    if logged_in?
+      erb :'/tweets/new'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+        erb :'tweets/edit'
+      else
+        redirect to '/tweets'
+      end
+    else
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id/edit' do
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  if params[:content] == ""
+    redirect "/tweets/#{params[:id]}/edit"
+  else
+    @tweet = Tweet.find(params[:id])
```
Lukeghenco: there is no need to set `@tweet` into an instance variable as this will not be in the view

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+    @tweet.update(content: params[:content])
+    redirect "/tweets/#{@tweet.id}"
+  end
+end
+
+delete '/tweets/:id/delete' do
+  @tweet = Tweet.find(params[:id])
```
Lukeghenco: see comment on line 56

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  get '/signup' do
+    if logged_in?
+      redirect '/tweets'
+    else
+      erb :'/users/signup'
+    end
+  end
+
+  post '/signup' do
+  if params[:username] == "" || params[:email] == "" || params[:password] == ""
```
Lukeghenco: You are already using validations on the model so we don't need to check for this now.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    redirect '/signup'
+  else
+    user = User.create(params)
```
Lukeghenco: I would do something like this to verify that user save to the database 

```ruby
user = User.new(params)
if user.save 
  session[:user_id] = user.id 
  redirect '/tweets'
else 
  redirect '/signup'
end
```

PR #224, opened 2016-10-26
Finished

radditude:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+    if logged_in
+      erb :'/tweets/new'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    if logged_in
+      @user = current_user
```
Lukeghenco: current_user is already saved as an instance variable from the parent class that we are inheriting from, so this just sets two instance variables.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+      
+      if params[:content].length <= 140
+        
+        @tweet = Tweet.new(content: params[:content], user_id: @user.id)
```
Lukeghenco: `current_user.tweets.build(params)` is the best way to do this. It creates an associative relationship for us already.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+      end
+      
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets' do
+    if logged_in
+      @user = current_user
```
Lukeghenco: see comment on line 13

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+      @tweets = Tweet.all
+      erb :'/tweets/index'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id' do
+    if logged_in
+      @user = current_user
```
Lukeghenco: see comment on line 13

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+      end
+      
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/:id' do
+    if logged_in
+      @user = current_user
```
Lukeghenco: see comment on line 13

app/models/user.rb
```diff
@@ -0,0 +1,15 @@
+class User < ActiveRecord::Base
+  has_many :tweets
+  has_secure_password
+  validates_presence_of :username, :email
```
Lukeghenco: you should also check for uniqueness of email and username

radditude:
Whoops, that was a rookie mistake! I fixed the double instance variable problems and the other things you pointed out. Thanks!

PR #225, opened 2016-10-26
Done

lynch16:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,61 @@
...
   end
 
-end
\ No newline at end of file
+  get '/' do
+      erb :index
+  end
+
+  get '/signup' do
+      if !!session[:id]
```
Lukeghenco: we should build out some helper methods in our `ApplicationController` that check for `current_user` and `logged_in?`

```ruby 

helpers do 

  def logged_in?
    !!current_user 
  end 

  def current_user 
  end 

end 
```

anytime we are now wanting access to the logged_in user just call `current_user` in your controllers and view(.erb).

And instead of checking session[:id] call `logged_in?`

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,61 @@
...
+          erb :signup
+      end
+  end
+
+  post '/signup' do
+      @user = User.new(params)
+      if !@user.save
+          redirect '/signup'
+      else
+          session[:id] = @user.id
```
Lukeghenco: session should be set after a successful save. 

```ruby 
post '/signup' do 
  user = User.new(params)
  if user.save 
    session[:id] = user.id 
    redirect '/tweets'
  else 
    redirect '/signup'
  end 
end 
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,61 @@
...
+          @user.save
+          redirect '/tweets'
+      end
+  end
+
+  get '/login' do
+      if !!session[:id]
+          @user = User.find(session[:id])
```
Lukeghenco: why are we looking for a user here. won't `get '/tweets'` do that for us?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,74 @@
+class TweetController < ApplicationController
+
+    get '/tweets' do
+        if !!session[:id]
+            @user = User.find(session[:id])
```
Lukeghenco: we can remove all of these `@user = User.find(session[:id])` since we have the `current_user` method now.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+            erb :'/tweets/new'
+        else
+            redirect '/login'
+        end
+    end
+
+    post '/tweets' do
+        @user = User.find(params[:user_id])
+        @tweet = Tweet.new(params)
```
Lukeghenco: we should check to make sure that a user is` logged_in?` to do this.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+        if !!session[:id]
+            @user = User.find(session[:id])
+            @tweet = Tweet.find(params[:id])
+            erb :'/tweets/edit'
+        else
+            redirect '/login'
+        end
+    end
+
+    post '/tweets/:id' do
```
Lukeghenco: this should be a put/patch statement. You will need to do a hidden tag in the form that handles a `put or patch` method.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+        if params[:content].empty?
+            redirect "/tweets/#{@tweet.id}/edit"
+        else
+            @tweet.content = params[:content]
+            @tweet.save
+            redirect "/tweets/#{@tweet.id}"
+        end
+    end
+
+    post '/tweets/:id/delete' do
```
Lukeghenco: this should be a `delete` statement. You will need to do a hidden tag in the form that handles a `delete` method.

PR #226, opened 2016-10-28
Done

yeungn:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,23 @@ class ApplicationController < Sinatra::Base
...
+    enable :sessions
+    set :session_secret, "password_security"
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,23 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+
+    enable :sessions
+    set :session_secret, "password_security"
+  end
+
+  helpers do
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      redirect "/login"
+    else
+      @tweet = Tweet.find(params[:id])
+      if @tweet.user_id == current_user.id 
+        erb :'tweets/edit_tweet'      
+      end
+    end
+  end
+
+  patch '/tweets/:id' do
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  get '/tweets/new' do
+    if logged_in?
+      erb :'tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
```
Lukeghenco: see comment on line 48

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end
+  end
+
+  patch '/tweets/:id' do
+    if params[:content].empty?
+      # binding.pry
+      redirect "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find(params[:id])
```
Lukeghenco: I would change the flow of this some.

```ruby 
tweet = Tweet.find(params[:id])
if tweet.update(params)  # this returns a true or false if it successfully commits to database
else 
  redirect to "/tweets/#{params[:id]}/edit"
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      @tweet.content = params[:content]
+      @tweet.save
+      redirect "/tweets/#{@tweet.id}"
+    end
+  end
+
+  post '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by(params[:id])
```
Lukeghenco: be specific here `find_by_id(params[:id]` make sure that it is querying the right column

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  get '/signup' do
+    if logged_in?
+      redirect '/tweets'
+    else
+      erb :'users/create_user'
+    end
+  end
+
+  post '/signup' do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      redirect '/signup'
+    else
+      @user = User.create(username: params[:username], email: params[:email], password: params[:password])
```
Lukeghenco: you don't have to be this explicit due to mass assignment with Active Record.

`user = User.create(params)` will work fine.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    if logged_in?
+      redirect '/tweets'
+    else
+      erb :'/users/login'
+    end
+  end
+
+  post '/login' do
+    @user = User.find_by(username: params[:username])
+    if @user != nil && @user.authenticate(params[:password])
```
Lukeghenco: `if @user && @user.authenticate(params[:password])` will work as well. 

@user would return false if there is none found.

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,5 @@
+<h3>Welcome, <%= current_user.username %></h3>
+
+<% Tweet.all.each do |tweet| %>
```
Lukeghenco: MVC design pattern dictates that interactions with `models` should be done in the controller.

PR #227, opened 2016-10-28
Done

nadukrow:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+        @tweet.update(content: params[:content])
+        redirect "/tweets/#{@tweet.id}"
+      end
+    else
+      redirect "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
```
Lukeghenco: no need for an instance variable of `@tweet` here just set it to a be `tweet = Tweet.find_by_id(params[:id])` as we will not be using this in the view.

nadukrow:
Thanks for the comment. I've removed the instance variable and just set it equal to. You're right. I wouldn't need an instance variable for the view.

PR #228, opened 2016-11-01
Done

johnram528:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,146 @@ class ApplicationController < Sinatra::Base
...
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "unbreakable_password"
   end
 
+  get '/' do
+    erb :index
+  end
+
+  get '/signup' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,146 @@ class ApplicationController < Sinatra::Base
...
+    redirect '/tweets'
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,146 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/login' do
+    if logged_in? 
+      redirect '/tweets'
+    else
+      erb :'/users/login'
+    end  
+  end
+
+  post '/login' do
```
Lukeghenco: you should set up a verification gate here to make sure a User is not logged_in? before they create another account. Otherwise a user could make a user by changing the parameters in a request.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,146 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    erb :'/users/user_show'
+  end
+
+  get '/tweets' do
+    if logged_in?
+      @user = User.find(session[:user_id])
```
Lukeghenco: you should have access to your `current_user` method here.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,146 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  post '/tweets' do
+    if !logged_in?
+      redirect '/tweets/new'
+    elsif  params[:content].empty? 
+      redirect '/tweets/new'
+    else
+      user = current_user
+      tweet = Tweet.create(content: params[:content], user_id: user.id)
```
Lukeghenco: The best way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,146 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+  post '/tweets' do
+    if !logged_in?
+      redirect '/tweets/new'
+    elsif  params[:content].empty? 
+      redirect '/tweets/new'
+    else
+      user = current_user
```
Lukeghenco: current_user is already set to an instance variable, so this is a duplicated variable

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,146 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end 
+
+  patch '/tweets/:id' do
+    if params[:content] == ""
+      redirect "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find(params[:id])
+      @tweet.update(content: params[:content])
+      @tweet.save
```
Lukeghenco: @tweet.update already returns a `true` or `false` value and commits to the database, so `.save` is redundant.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,146 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in? 
+      redirect '/tweets'
+    else
+      erb :'/users/create_user'
+    end
+  end
+
+  post '/signup' do
+    user = User.new(params) 
+    if !user.attributes.values.include?("")&& user.save
```

PR #229, opened 2016-11-03
Done

kkterai:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,30 @@
-require './config/environment'
+require "./config/environment"
+require "./app/models/user"
+require "./app/models/tweet"
 
 class ApplicationController < Sinatra::Base
```
Lukeghenco: Great job splitting your actions among multiple controllers!

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,30 @@
...
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+  		def logged_in?
+  			!!session[:user_id]
+  		end
+
+  		def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,30 @@
...
-end
\ No newline at end of file
+#-------- Landing page --------
+
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+  		def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    if !logged_in?
+      redirect "/login"
+    else
+      erb :'tweets/create_tweet'
+    end
+  end
+
+  post '/tweets' do
+    if  logged_in?
+      @tweet = Tweet.new(:content => params["tweet"])
```
Lukeghenco: the best way to handle this would be to make use of the `has_many` relationship on User:
```
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  patch '/tweets/:id' do
+    if params["tweet"] == ""
+      @tweet = Tweet.find(params[:id])
+      redirect "/tweets/#{@tweet.id}/edit"
+    else
+      @tweet = Tweet.find(params[:id])
+      @tweet.update(:content => params["tweet"])
+      @tweet.save
```
Lukeghenco: @tweet.update already save to the database, so `.save` would be redundant at this point

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  get '/signup' do
+    if logged_in?
+      redirect '/tweets'
+    else
+      erb :'users/create_user'
+    end
+  end
+
+  post "/signup" do
+    user = User.new(username: params[:username], email: params[:email], password: params[:password])
```
Lukeghenco: I would check to make sure a user is logged_in? before you allow them to make a user. At the moment anyone could force create more users.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  get '/login' do
+    if !logged_in?
+      erb :'users/login'
+    else
+      redirect "/tweets"
+    end
+  end
+
+  post "/login" do
```
Lukeghenco: I would setup a `gate` here to make sure a user is not logged_in? before they can proceed

app/helpers/helpers.rb
```diff
@@ -0,0 +1,11 @@
+class Helpers
+
+  def self.current_user(session)
```
Lukeghenco: you already have these methods in your ApplicationController

app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,14 @@
+<% if logged_in?  %>
+       <h1>Welcome, <%= current_user.username %></h1>
+
+<%Tweet.all.each do |tweet| %>
```
Lukeghenco: With an MVC design pattern, it is best practice to separate the concerns of Model calls in the controller. Example:

```ruby 
get '/tweets' do 
  @tweets = Tweet.all
end
``` 

and now you will have access to `<% @tweets.each do |tweet| %>`

app/controllers/users_controller.rb
kkterai: Getting conflicting information from AAQ.

Lukeghenco:
```ruby 
post "/signup" do 
  if !logged_in?
    user = User.new(params)
    if user.save 
      redirect to '/tweets'
    else 
      redirect to '/signup'
     end 
   else 
     redirect to '/tweets' 
   end 
end
```

kkterai:
Fixes completed! :-)

PR #236, opened 2016-11-18
Project submitted

brettCole:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+  get '/signup' do
+    if !!session[:user_id]
+    	redirect '/tweets'
+    else
+    	erb :'/users/create_user'
+    end
+  end
+
+  post '/signup' do
+  	if params.any?{|key, value| value.empty?}
```
cernanb: I saw you used ActiveRecord validations in your Tweet model. You could have also use them here as well. 

[ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+
+  helpers do
+    def logged_in?
+      !!current_user
+    end
+
+    def current_user
+      @current_user ||= User.find(session[:id]) if session[:id]
```
cernanb: Great job on using memoization!

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      erb :'/tweets/create_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do
+    @tweet = Tweet.create(content: params[:content], user_id: session[:id])
+    @tweet.save
```
cernanb: Using `.create` instantiates and saves the object to the database...so there is no need to call `.save` in the following line. If you were to have done `Tweet.new`...then the `.save`. would have been necessary.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+         erb :'tweets/edit_tweet'
+       else
+         redirect "/tweets"
+       end
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets/:id' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+    @tweet = Tweet.find(params[:id])
+    @tweet.update(content: params[:content])
+    redirect "/tweets/#{session[:id]}/edit"
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #238, opened 2016-11-23
Done

pianistps:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "secret"
   end
 
-end
\ No newline at end of file
+  helpers do
+    def current_user
+      @user = User.find_by_id(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+    end
+
+    def is_logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  get '/tweets/new' do
+    if is_logged_in?
+      erb :'/tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] == ""
```
cernanb: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
+require './config/environment'
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+      erb :'tweets/show_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if is_logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      erb :'/tweets/edit_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/:id' do
+    @tweet = Tweet.find_by_id(params[:id])
+    if params[:content] != ""
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  get '/tweets/:id/edit' do
+    if is_logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      erb :'/tweets/edit_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/:id' do
```
cernanb: Consider using the `patch` http verb when updating the tweet.

app/controllers/application_controller.rb
pianistps: Yes, makes sense! Thanks! Is this what you were suggesting? 
`  def current_user
      @user ||= User.find_by_id(session[:user_id]) if session[:user_id]
    end`

app/controllers/tweets_controller.rb
pianistps: Thanks!

app/controllers/tweets_controller.rb
pianistps: fixed

app/controllers/tweets_controller.rb
pianistps: that makes sense!

PR #243, opened 2016-12-06
Done

em379:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,144 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "secret"
   end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,144 @@ class ApplicationController < Sinatra::Base
...
+    erb :index
+  end
+
+  get '/users/:slug' do
+      @user = User.find_by_slug(params[:slug])
+      erb :'/users/show'
+    end
+
+  get '/signup' do
+    if session[:user_id]
```
cernanb: You can use your `logged_in?` helper method here, which is preferable to referring to the session object directly.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,144 @@ class ApplicationController < Sinatra::Base
...
+      session[:user_id] = @user.id
+      redirect '/tweets'
+    else
+      redirect '/signup'
+    end
+  end
+
+  get '/tweets' do
+    if session[:user_id]
+      @user = User.find(session[:user_id])
```
cernanb: Your `current_user` method could be useful here.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,144 @@ class ApplicationController < Sinatra::Base
...
+    if session[:user_id]
+      redirect '/tweets'
+    else
+      erb :'/users/create_user'
+    end
+  end
+
+  post '/signup' do
+    @user = User.new(username: params["username"], email: params["email"], password: params["password"])
+    if !params[:username].empty? && !params[:email].empty? && @user.save
```
cernanb: I think you could use [ActiveRecord validations](https://learn.co/tracks/full-stack-web-development/sinatra/activerecord/user-authentication-in-sinatra) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,144 @@ class ApplicationController < Sinatra::Base
...
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if !session[:user_id]
+      redirect to '/'
+    elsif !params[:content].empty?
+      @user = User.find(session[:user_id])
+      @tweet = Tweet.create(content: params[:content])
```

``` ruby
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,144 @@ class ApplicationController < Sinatra::Base
...
+      erb :'/tweets/show_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:tweet_id/edit' do
+    if session[:user_id]
+      @user = User.find(session[:user_id])
+      @tweet = Tweet.find(params[:tweet_id])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,144 @@ class ApplicationController < Sinatra::Base
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:tweet_id' do
+    @tweet = Tweet.find(params[:tweet_id])
+    if !session[:user_id]
+      redirect to '/'
+    elsif session[:user_id] != @tweet.user.id
```
cernanb: Great job checking to see if the tweet belongs to the logged in user before updating it!

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,144 @@ class ApplicationController < Sinatra::Base
...
+      redirect to "/tweets/#{@tweet.id}/edit"
+    else
+      @tweet.update(content: params[:content])
+      @tweet.save
+
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:tweet_id' do
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #248, opened 2016-12-10
Project submitted

lmilne360:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
+class TweetsController < ApplicationController
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+		if logged_in?
+			erb :'tweets/new'
+		else
+			redirect '/login'
+		end
+	end
+
+	post '/tweets' do
+		if params[:content].blank?
+			redirect "/tweets/new"
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,24 @@
...
+  get '/' do
+  	erb :index
+  end
+
+  helpers do	
+		def logged_in?
+			!!current_user
+		end
+
+		def current_user
```
cernanb: Excellent job memoizing current_user!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+		 	else
+		 		redirect to '/tweets'
+		 	end
+		
+		else
+			redirect '/login'
+		end
+	end
+
+	post '/tweets/:id' do
```
cernanb: Consider using `patch` for the http method. 

```
patch '/tweets/:id' do
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+		else
+			@tweet = Tweet.find_by_id(params[:id])
+			@tweet.update(content: params[:content])
+			redirect "/tweets/#{@tweet.id}"
+		end		
+	end
+
+	post '/tweets/:id/delete' do 
+		if logged_in?
+			@tweet = Tweet.find_by_id(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

lmilne360:
Hey Cern, I just pushed the changes you suggested.

PR #249, opened 2016-12-10
Done

christinechoi:


app/controllers/application_controller.rb
```diff
@@ -1,4 +1,5 @@
 require './config/environment'
+require 'pry'
 
 class ApplicationController < Sinatra::Base
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +8,122 @@ class ApplicationController < Sinatra::Base
     set :views, 'app/views'
   end
 
+  enable :sessions
+
```
cernanb: Consider creating helper methods to determine if a user is logged in and to grab user if so.

```
def logged_in?

end

def current_user

end
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +8,122 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/new' do 
+    redirect :'/login' if !session[:id]
+    erb "<form action='/tweets/new' method='post'>
+    <input type='text' name='content'>Tweet</input>
+    <input type='submit' id='submit'>
+    </form>"
+  end
+
+  post '/tweets/new' do 
+    redirect :'/tweets/new' if params[:content].blank?
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +8,122 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/tweets/new' do 
+    redirect :'/login' if !session[:id]
+    erb "<form action='/tweets/new' method='post'>
+    <input type='text' name='content'>Tweet</input>
+    <input type='submit' id='submit'>
+    </form>"
+  end
+
+  post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +8,122 @@ class ApplicationController < Sinatra::Base
...
+  get "/users/:slug" do 
+    @user = User.find_by_slug(params[:slug])
+    erb "
+      <%= @user.tweets.each {|tweet| tweet.content} %> 
+    "
+  end
+
+  get "/tweets/:id/edit" do 
+    redirect :'/login' if !session[:id]
+    @tweet = Tweet.find(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +8,122 @@ class ApplicationController < Sinatra::Base
...
+    session[:id] = user.id
+
+    redirect :'/tweets'
+  end
+
+  get '/login' do
+
+    redirect :'/tweets' if session[:id]
+    
+    erb "<%= session[:id].nil? %>
```
cernanb: It's normally best practice to render an erb file rather than have your markup in the controller.

app/controllers/application_controller.rb
```diff
@@ -7,4 +8,122 @@ class ApplicationController < Sinatra::Base
...
+    @tweet = Tweet.find(params[:id])
+
+    erb "<form action='/tweets/edit' method='post'>
+    <input type='text' name='content' value='<%= @tweet.content %>'>Tweet</input>
+    <input type='text' name='id' value=<%= @tweet.id %>></input>
+    <input type='submit' id='submit'>
+    </form>"
+  end
+
+  post "/tweets/edit" do
```
cernanb: The convention for the update action is `patch 'tweets/:id' do `.

PR #253, opened 2016-12-15
Done

storywr:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,139 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "secret"
   end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,139 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/new' do
+    if logged_in?
+      erb :'tweets/create_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] != ""
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,139 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,139 @@ class ApplicationController < Sinatra::Base
...
+      session.clear
+      redirect "/login"
+    else
+      redirect "/login"
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
```
cernanb:
```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,139 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      erb :'tweets/create_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] != ""
+      @tweet = Tweet.new(content: params[:content], user_id: current_user.id)
```
```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,139 @@ class ApplicationController < Sinatra::Base
...
+      @tweet = Tweet.find_by(params[:id])
+      erb :'tweets/show_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,139 @@ class ApplicationController < Sinatra::Base
...
+      erb :'tweets/edit_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find_by(params[:id])
+    if params[:content] != ""
+      @tweet.content = params[:content]
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,139 @@ class ApplicationController < Sinatra::Base
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    else
+      redirect to "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by(params[:id])
```
cernanb: Here is another place you would want to ensure that the tweet that is being deleted is associated with the current user. 

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #254, opened 2016-12-16
Project submitted

adamfriedl:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    else
+      redirect :"/tweets/#{@tweet.id}"
+    end
+  end
+
+  patch '/tweets/:id' do
+    if params[:content] == ""
+      redirect "tweets/params[:id]/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
```

PR #258, opened 2016-12-29
Done

kjleitz:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
+class TweetsController < Sinatra::Base
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    @tweets = Tweet.all
+    erb :tweets
+  end
+
+  get "/tweets/new" do
+    erb :new
+  end
+
+  post "/tweets" do
+    if params[:tweet][:content].empty?
```

app/controllers/concerns/helper.rb
```diff
@@ -0,0 +1,11 @@
+module Helper
+
+  def logged_in?
+    !!session[:id]
+  end
+
+  def current_user
+    User.find(session[:id])
```
cernanb: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/controllers/concerns/helper.rb
```diff
@@ -0,0 +1,11 @@
+module Helper
+
+  def logged_in?
+    !!session[:id]
```
cernanb: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      flash[:message] = "Tweet contents cannot be empty!"
+      redirect to("/tweets/new")
+    else
+      current_user.tweets.create(params[:tweet])
```
cernanb: Awesome job taking advantage of the has_many relationship to create a tweet!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+  get "/tweets/:id" do
+    @user = current_user
+    @tweet = Tweet.find(params[:id])
+    erb :show_tweet
+  end
+
+  get "/tweets/:id/edit" do
+    @tweet = Tweet.find(params[:id])
+    if @tweet.user == current_user
+      erb :edit_tweet
```
cernanb: I don't see a lot of students checking the tweet's user. Good job!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    else
+      flash[:message] = "This is not your tweet to edit!"
+    end
+    redirect to("/tweets/#{tweet.id}")
+  end
+
+  delete "/tweets/:id/delete" do
+    tweet = Tweet.find(params[:id])
+    if tweet.user == current_user
+      tweet.destroy
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/tweets_controller.rb
kjleitz: OH! I had forgotten about those. Totally right! I'll fix that.

app/controllers/tweets_controller.rb
kjleitz: That's a really good point about not manually checking. Also, such a good idea `destroy`ing in the `if` statement! I'll definitely change that.

Also, I did NOT know that about `find_by` versus `find`! I naively assumed that `find` was an alias for `find_by(id: ...)` (in classic Ruby "more ways than one" style)! That's very good to know. I'll have to read up on the documentation of those methods.

app/controllers/concerns/helper.rb
kjleitz: I assume "memoized" means that it doesn't re-run the method when it's called successively, sort of like a cached value? I'll look into that, good to know! I'll change this too, and make a note for the future--you're totally right, even just in the fact that I should use abstraction whenever possible/appropriate.

app/controllers/concerns/helper.rb
kjleitz: Ohhhh, damn that makes a lot of sense. I bet that's a _hell_ of a lot more efficient than database lookups. Can't wait to check it out.

PR #259, opened 2017-01-11
Done

jonbf:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,110 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "password_security"
   end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,110 @@ class ApplicationController < Sinatra::Base
...
+    if !logged_in? then redirect "/login" end
+    @tweet = Tweet.find(params[:id].to_i)
+    @user=@tweet.user
+    if session[:id]!= @user.id then redirect "/tweets" end
+    erb :"/tweets/edit_tweet"
+  end
+
+  post '/tweets/new' do
+    if !logged_in? then redirect "/login" end
+    if params[:content]=="" then redirect "/tweets/new" end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,110 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+   helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,110 @@ class ApplicationController < Sinatra::Base
...
+      session[:id]=user.id
+      redirect "/tweets"
+    else
+      redirect "/login"
+    end
+  end
+
+   helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,110 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/tweets/:id/edit' do
+    if !logged_in? then redirect "/login" end
+    @tweet = Tweet.find(params[:id].to_i)
+    @user=@tweet.user
+    if session[:id]!= @user.id then redirect "/tweets" end
+    erb :"/tweets/edit_tweet"
+  end
+
+  post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,110 @@ class ApplicationController < Sinatra::Base
...
+    if !logged_in? then redirect "/login" end
+    if params[:content]=="" then redirect "/tweets/new" end
+    @tweet = Tweet.create(content: params[:content], user_id: session[:id])
```
```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,110 @@ class ApplicationController < Sinatra::Base
...
+   @session=session
+    @tweet = Tweet.find(params[:id])
+    erb :"/tweets/show_tweet"
+  end
+
+  get '/tweets/:id/edit' do
+    if !logged_in? then redirect "/login" end
+    @tweet = Tweet.find(params[:id].to_i)
+    @user=@tweet.user
+    if session[:id]!= @user.id then redirect "/tweets" end
```
cernanb: Great work confirming the tweet belongs to the user that's signed in before attempting an edit.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,110 @@ class ApplicationController < Sinatra::Base
...
+  post '/tweets/new' do
+    if !logged_in? then redirect "/login" end
+    if params[:content]=="" then redirect "/tweets/new" end
+    @tweet = Tweet.create(content: params[:content], user_id: session[:id])
+    redirect "/tweets"
+  end
+
+  post '/tweets/:id' do
+    if params[:content]=="" then redirect "/tweets/#{params[:id]}/edit" end
+    @tweet=Tweet.find(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,110 @@ class ApplicationController < Sinatra::Base
...
+    @tweet.update(content: params[:content])
+    @tweet.save
+    redirect "/tweets"
+  end
+
+  post '/tweets/:id/delete' do
+    tweet=Tweet.find(params[:id])
+    if tweet.user.id==session[:id] then tweet.delete end
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #260, opened 2017-01-16
Project submitted

aml7733:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,150 @@
 require './config/environment'
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,150 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    if !logged_in?
+      redirect '/login'
+    else
+      Tweet.create(content: params[:content], user_id: session[:id]) unless params[:content] == ""
```

Also another way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,150 @@
...
+  end
+
+
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+
+    def current_user
+      User.find(session[:id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,150 @@
...
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    erb :'/tweets/tweets'
+  end
+
+
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,150 @@
...
+  get '/tweets/new' do
+    if logged_in?
+      erb :'/tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    if !logged_in?
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,150 @@
...
+      @tweet = Tweet.find(params[:id])
+      erb :'/tweets/show_tweet'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if !logged_in?
+      redirect '/login'
+    else
+      @tweet = Tweet.find(params[:id])
```
cernanb: Consider verifying a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit.

PR #261, opened 2017-01-16
Done

krishl:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
+class TweetsController < ApplicationController
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    if logged_in?
+      @tweets = Tweet.all
+      erb :'tweets/tweets'
+    else
+      redirect to "/login"
+    end
+  end
+
+  post '/tweets' do
+    if params[:content].empty?
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+    erb :index
+  end
+
+  helpers do
+    def logged_in?
+      !!current_user
+    end
+
+    def current_user
+      @current_user ||= User.find(session[:user_id]) if session[:user_id]
```
cernanb: Awesome job memoizing the current user!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      redirect to "/tweets/new"
+    else
+      @tweet = Tweet.create(content: params[:content], user_id: current_user.id)
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      erb :'tweets/show_tweet'
+    else
+      redirect to "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == current_user.id
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      redirect to "/login"
+    end
+  end
+
+  patch '/tweets/:id' do
+    if params[:content].empty?
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params[:content]
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      @tweet.save
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
```
cernanb: Consider finding tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #262, opened 2017-01-16
Done

senhad:


app/models/user.rb
```diff
@@ -0,0 +1,15 @@
+class User < ActiveRecord::Base 
+	has_many :tweets
+	has_secure_password
+	validates_presence_of :username, :email
```
cernanb: Great job using Activerecord validations!

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,77 @@ class ApplicationController < Sinatra::Base
...
+
+
+	helpers do 
+
+	  def logged_in?
+	  	!!current_user
+	  end
+
+	  def current_user
+	  	@current_user ||= User.find(session[:id]) if session[:id]
```
cernanb: Awesome work memoizing the current user!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+		@tweet = Tweet.find_by_id(params[:id])
+		erb :'tweets/show_tweet'
+	else 
+		redirect '/login'
+	end 
+end 
+
+get '/tweets/:id/edit' do 
+	if logged_in?
+		@tweet = Tweet.find_by_id(params[:id])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+		erb :'tweets/edit_tweet'
+	else 
+		redirect '/login'
+	end 
+end 
+
+patch '/tweets/:id' do 
+	@tweet = Tweet.find_by_id(params[:id])
+	if params[:content] == ""
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+		redirect "/tweets/#{@tweet.id}/edit"
+	else 
+		@tweet.content = params[:content]
+		@tweet.save
+		redirect "/tweets/#{@tweet.id}"
+	end 
+end 
+
+delete '/tweets/:id/delete' do
+	@tweet = Tweet.find_by_id(params[:id])
```
cernanb: Consider finding tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #263, opened 2017-01-17
Done

howardbdev:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,58 @@
+class TweetsController < ApplicationController
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  get '/tweets/new' do
+    if logged_in?
+      erb :'/tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    redirect '/tweets/new' if params[:content].empty?
```

```
@tweet = Tweet.new(content: params[:content], user_id: current_user.id)
if @tweet.save
  erb :'/tweets/show_tweet'
else
  redirect '/tweets/new'
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,39 @@ class ApplicationController < Sinatra::Base
...
+      if user && user.authenticate(password)
+        session[:user_id] = user.id
+        redirect '/tweets'
+      else
+        redirect '/login'
+      end
+    end
+
+    def current_user
+      @current_user ||= User.find_by(id: session[:user_id])
```
cernanb: Great job using memoization for current_user!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    @tweet = Tweet.new(content: params[:content], user_id: current_user.id)
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+      erb :'/tweets/show_tweet'
+    else
+      redirect '/tweets'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    redirect '/login' unless logged_in?
+    @tweet = Tweet.find_by_id(params[:id])
+    if @tweet
```
cernanb: You would probably want to verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+      erb :'/tweets/edit_tweet'
+    else
+      redirect '/tweets'
+    end
+  end
+
+  patch '/tweets/:id/edit' do
+    @tweet = Tweet.find_by_id(params[:id])
+    @tweet.content = params[:content] unless params[:content].empty?
```
cernanb: You would probably also want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    @tweet.save
+    erb :'/tweets/show_tweet'
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
+    @tweet.delete if @tweet.user_id == current_user.id
```
cernanb: Great job checking to make sure the tween belongs to the current user here! 

One thing to consider here is finding tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

howardbdev:
Thanks Cernan!

I went back through and refactored, I believe I hit all your suggestions.

BTW I never got any notifications from git.  Is that something I should be
concerned about?  The only place I've found your comments are in this email.

Please let me know my next step.  I think it's to submit a project idea to
you for the final Sinatra portfolio project, is that correct?

Thanks again
Howard

PR #264, opened 2017-01-19
Done

gd875:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
   configure do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+     if logged_in?
+         erb :'/tweets/create_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do
+    @user = User.find(session[:user_id])
+    if !params["content"].empty?
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+        end
+    end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+                redirect "/tweets"
+            end
+        else
+          redirect "/login"
+        end
+    end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+         erb :'/tweets/create_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do
+    @user = User.find(session[:user_id])
+    if !params["content"].empty?
+        @user.tweets << Tweet.create(content: params["content"])
```
```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+      redirect "/login"
+    end
+  end
+
+    get '/tweets/:tweet_id/edit' do
+        if logged_in?
+            @user = User.find(session[:user_id])
+            @tweet_id = params[:tweet_id]
+            @tweet = Tweet.find_by(id: @tweet_id)
+            if @tweet.user_id == @user.id
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+        else
+          redirect "/login"
+        end
+    end
+
+    patch '/tweets/:tweet_id' do
+        @tweet_id = params[:tweet_id]
+        @tweet = Tweet.find_by(id: @tweet_id)
+        if !params["content"].empty?
+            @tweet.update(content: params["content"])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,155 @@ class ApplicationController < Sinatra::Base
...
+            redirect to '/tweets'
+        else
+            redirect to "/tweets/#{@tweet_id}/edit"
+        end
+    end
+
+    delete '/tweets/:tweet_id/delete' do
+        if logged_in?
+            @user = User.find(session[:user_id])
```
cernanb: You can probably take advantage of your `current_user` helper method here.

PR #266, opened 2017-01-23
Done

christinalcole:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
+require 'twilio-ruby'
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+  get '/tweets/new' do
+    if Helpers.is_logged_in?(session)
+      erb :'/tweets/new'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content].empty?
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,9 @@
+class Helpers
+  def self.current_user(session_hash) # uses the user_id from the session hash to find the user in the database and return that user.
+    @user = User.find_by_id(session_hash[:user_id])
```
cernanb: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,9 @@
...
+  end
+
+  def self.is_logged_in?(session_hash)  # returns true if the user_id is in the session hash and false if not.
+    !!session_hash[:user_id]
```
cernanb: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+      redirect to '/tweets/new'
+    else
+      @tweet = Tweet.create(content: params[:content], user_id: session[:user_id])
+      # if params[:phone_number] != ""
```
cernanb: Another way to do this is to make use of the `has_many` relationship on User that we discussed recently:
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+      erb :'/tweets/show'
+    end
+  end
+
+  get '/tweets/:id/edit' do #tweet edit page
+    if !Helpers.is_logged_in?(session)
+      redirect to '/login'
+    else
+    @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == session[:user_id]
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+        redirect to '/tweets'
+      end
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find_by_id(params[:id])
+
+    if !params[:content].empty?
+      @tweet.update(content: params[:content])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+      redirect to "/tweets/#{@tweet.id}"
+    else
+      redirect to "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  delete '/tweets/:id/delete' do #delete tweet action
+    @tweet = Tweet.find_by_id(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

Excellent usage of `find_by_id`!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  get '/tweets/:id' do #tweet show page
+    if !Helpers.is_logged_in?(session)
+      redirect to '/login'
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      erb :'/tweets/show'
+    end
+  end
+
```
cernanb: Test Comment

christinalcole: thx!  notifications problem resolved!

app/controllers/tweets_controller.rb
christinalcole: refactored to validate presence of tweet content...not 100% that I caught all instances, but this one, at least, is updated

app/controllers/tweets_controller.rb
christinalcole: This goes along with a comment 2 down: associating @tweet with the current_user, as you've suggested, would be a more robust (? better word?) way to prevent edits from other users?

app/controllers/tweets_controller.rb
christinalcole: then, for example, this would be a good place to use AR's has_many method, to associate the user with the tweet on creation?  then verify that @tweet.user = current_user?

PR #271, opened 2017-01-31
Done

mo6709:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,80 @@
+class TweetController < ApplicationController
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+  	if logged_in?
+  	  erb :'tweets/new'
+  	else
+  		redirect to '/login'
+  	end  
+  end
+
+  post '/tweets' do
+    @user = User.find_by_id(session[:user_id])
+    if params[:content] == ""
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,23 @@ class ApplicationController < Sinatra::Base
...
+  end
+  
+
+  helpers do
+    def logged_in?
+       !!session[:user_id]
+    end 
+
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,23 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  get '/' do
+    erb :ftweeter_home_page
+  end
+  
+
+  helpers do
+    def logged_in?
+       !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+  get '/tweets/new' do
+  	if logged_in?
+  	  erb :'tweets/new'
+  	else
+  		redirect to '/login'
+  	end  
+  end
+
+  post '/tweets' do
+    @user = User.find_by_id(session[:user_id])
```
cernanb: You can take advantage of your `current_user` helper here. 

`@user = current_user`

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+	    @tweet = Tweet.find_by_id(params[:id])
+	  	erb :'tweets/show'
+	  else
+	  	redirect '/login'
+	  end
+  end
+
+  get '/tweets/:id/edit' do
+  	if logged_in?
+  		@tweet = Tweet.find_by_id(params[:id])
```
cernanb: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+  		erb :'tweets/edit'
+  	else
+  		redirect '/login'
+  	end	
+  end
+
+  post '/tweets/:id' do
```
cernanb: Convention would be to use the `patch` http verb here instead of `post`.

PR #279, opened 2017-02-12
Project submitted

jilustrisimo:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,89 @@
+class TweetsController < ApplicationController
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+        redirect to '/tweets'
+      end
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params.value?('')
+      flash[:notice] = 'Cannot create empty tweet.'
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,27 @@
...
+    erb :index
+  end
+
+  helpers do
+
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+      redirect to '/tweets/new'
+    else
+      user = User.find(session[:user_id])
+      Tweet.create(user_id: user.id, content: params[:content])
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+    else
+      flash[:notice] = 'Please log in first to view tweets.'
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == current_user.id
```

PR #292, opened 2017-03-01
Project submitted

kevinladkins:


app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
 class ApplicationController < Sinatra::Base
 
   configure do
+    enable :sessions
+    set :session_secret, "pollywog"
     set :public_folder, 'public'
     set :views, 'app/views'
   end
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
...
+  get '/tweets/new' do
+    if !logged_in?
+      redirect '/login'
+    else
+      erb :'tweets/create_tweet'
+    end
+  end
+
+  post '/tweets/new' do
+    if params[:content] == ""
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
...
+  end
+
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
...
+  end
+
+  post '/tweets/:id/delete' do
+    redirect "/tweets/#{params[:id]}/delete"
+  end
+
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
...
+
+  get '/tweets/new' do
+    if !logged_in?
+      redirect '/login'
+    else
+      erb :'tweets/create_tweet'
+    end
+  end
+
+  post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
...
+  get '/tweets/:id/edit' do
+    if !logged_in?
+      redirect "/login"
+    else
+      @tweet = Tweet.find(params[:id])
+      erb :'tweets/edit_tweet'
+    end
+  end
+
+  post '/tweets/:id/edit' do
```
cernanb: Typically an update would be a `patch` request and the route would be to `/tweets/:id`

```
patch '/tweets/:id'
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
...
+    else
+      tweet = Tweet.find(params[:id])
+      tweet.content = params[:content]
+      tweet.save
+      redirect "/tweets/#{tweet.id}"
+    end
+  end
+
+  get '/tweets/:id/delete' do
+    if !logged_in?
```
cernanb: Typically a delete would be a delete request to `/tweets/:id`

```
delete '/tweets/:id'
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
...
+    else
+      erb :'tweets/create_tweet'
+    end
+  end
+
+  post '/tweets/new' do
+    if params[:content] == ""
+      redirect '/tweets/new'
+    else
+    @tweet = Tweet.create(content: params[:content], user_id: current_user.id)
```
```
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
...
+      @tweet = Tweet.find(params[:id])
+      erb :'tweets/show_tweet'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if !logged_in?
+      redirect "/login"
+    else
+      @tweet = Tweet.find(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,153 @@
...
+    end
+  end
+
+  post '/tweets/:id/edit' do
+    if !logged_in?
+      redirect "/login"
+    elsif params[:content] == ""
+      redirect "tweets/#{params[:id]}/edit"
+    else
+      tweet = Tweet.find(params[:id])
```

PR #301, opened 2017-03-20
Done

mdo5004:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
...
+
+    get "/tweets/:id/edit" do
+        if logged_in? && current_user==Tweet.find(params[:id]).user.id
+            @tweet = Tweet.find(params[:id])
+            erb :"tweets/edit_tweet"
+        else
+            redirect '/login'
+        end
+    end
+    post "/tweets/:id/edit" do
```
cernanb: Convention would be to use the `patch` http verb for this route.

```ruby
patch '/tweets/:id' do

end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
...
+        @tweet = Tweet.find(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,145 @@
...
+            redirect '/login'
+        end
+    end
+
+    get "/logout" do
+        session.clear
+        redirect "/login"
+    end
+    post "/tweets/:id/delete" do
+        @tweet = Tweet.find(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #302, opened 2017-03-26
Done

bentonwong:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      session.clear
+      redirect to '/login'
+    else
+      redirect to '/'
+    end
+  end
+
+  helpers do
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+    logged_in? ? (erb :'tweets/create_tweet') : (redirect to '/login')
+  end
+
+  post '/tweets' do
+    if logged_in? && (current_user.id == session[:id])
+      if params[:content].strip.empty?
+        redirect to '/tweets/new'
+      else
+        params[:user_id] = current_user.id
+        @tweet = Tweet.create(params)
```

``` ruby
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+      !!@tweet ? (erb :'tweets/show_tweet') : (redirect to '/tweets')
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    @tweet = Tweet.find_by_id(params[:id])
+    if logged_in?
+      if current_user.id == @tweet.user_id
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+      end
+    else
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    if params[:content].strip.empty?
+      redirect to "/tweets/#{params[:id]}/edit"
+    else
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+        @tweet = Tweet.find_by_id(params[:id])
+        @tweet.content = params[:content]
+        @tweet.save
+      end
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #306, opened 2017-04-05
Done

theschubinator:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  		if !!session[:id]
+  			erb :"tweets/create_tweet"
+  		else
+  			redirect "/login"
+  		end
+  	end
+
+  	post "/tweets" do 
+  		user = User.all.find(session["id"])
+		if params[:content] == ""
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,75 @@
+class TweetController < ApplicationController
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  			redirect "/login"
+  		end
+  	end
+
+  	post "/tweets" do 
+  		user = User.all.find(session["id"])
+		if params[:content] == ""
+			redirect "/tweets/new"
+		else
+			user.tweets << Tweet.create(params)
```
cernanb: Another way you can create an associated tweet is like so:

``` ruby
Tweet.create(content: params[:content], user: user)

File: app/controllers/tweet_controller.rb [hunk 4]
```diff
@@ -0,0 +1,75 @@
...
+	  		@tweet = Tweet.find(params[:id])
+	  		erb :"tweets/show_tweet"
+	  	else
+	  		redirect "/login"
+	  	end
+  	end
+
+  	get "/tweets/:id/edit" do
+  		if !!session[:id]
+  			@tweet = Tweet.find(params[:id])
```
cernanb: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  			erb :"tweets/edit_tweet"
+  		else
+  			redirect "/login"
+  		end
+  	end
+
+  	post "/tweets/:id" do
+  		tweet = Tweet.find(params[:id])
+  		if params[:content] == ""
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  			redirect "/tweets/#{tweet.id}/edit"
+  		else
+	  		tweet.content = params[:content]
+	  		tweet.save
+	  		redirect "/tweets/#{tweet.id}"
+	  	end
+  	end
+
+  	post "/tweets/:id/delete" do
+  		tweet = Tweet.find(params[:id])
```
cernanb: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  		if params[:content] == ""
+  			redirect "/tweets/#{tweet.id}/edit"
+  		else
+	  		tweet.content = params[:content]
+	  		tweet.save
+	  		redirect "/tweets/#{tweet.id}"
+	  	end
+  	end
+
+  	post "/tweets/:id/delete" do
```
cernanb: Convention for this route would be using the `delete` http verb.

PR #307, opened 2017-04-05
Done

walwoodr:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,137 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions unless test?
+    set :session_secret, "secret"
   end
```

app/models/tweet.rb
```diff
@@ -0,0 +1,5 @@
+class Tweet < ActiveRecord::Base
+  belongs_to :user
+  validates_presence_of :content, :user_id
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,137 @@ class ApplicationController < Sinatra::Base
...
+    end
+    redirect '/tweets'
+  end
+
+  helpers do
+   def is_logged_in?
+     !!session[:user_id]
+   end
+
+   def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,137 @@ class ApplicationController < Sinatra::Base
...
+    @tweet = Tweet.find(params[:id])
+    if is_logged_in? && current_user == @tweet.user
+
+      @tweet.destroy
+    end
+    redirect '/tweets'
+  end
+
+  helpers do
+   def is_logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,137 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/new' do
+    if is_logged_in?
+      erb :'tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    @user = User.find(session[:user_id])
```
cernanb: You can use your current_user method here and build a tweet associated with that user
`@tweet = current_user.tweets.build(params)`

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,137 @@ class ApplicationController < Sinatra::Base
...
+    @tweet = @user.tweets.build(params)
+    if @tweet.valid?
```
cernanb: You can also attempt to save in your if condition
```
if @tweet.save
    redirect "/tweets/#{@tweet.id}"
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,137 @@ class ApplicationController < Sinatra::Base
...
+    if params[:content] != ""
+      @tweet.update(content: params[:content])
+      redirect "/tweets/#{@tweet.id}"
+    else
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  delete '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #312, opened 2017-04-10
Project submitted

danpaulgo:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,90 @@
+class TweetsController < ApplicationController
```

app/models/tweet.rb
```diff
@@ -0,0 +1,6 @@
+class Tweet < ActiveRecord::Base
+  belongs_to :user
+  validates :content, presence: true
```

app/helpers/user_helper.rb
```diff
@@ -0,0 +1,11 @@
+class UserHelper
+
+  def self.current_user(session)
```
cernanb: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+      @flash_message = session[:flash] if session[:flash]
+      session[:flash] = nil
+      erb :'/tweets/new'
+    else
+      redirect "/login"
+    end
+  end
+
+  post "/tweets" do
+    @tweet = Tweet.new(params)
```
```
@tweet = UserHelper.current_user.tweets.build(content: params[:content])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+        erb :'/tweets/edit'
+      else
+        redirect "/tweets"
+      end
+    end
+
+  end
+
+  patch "/tweets/:id" do
+    @tweet = Tweet.find(params[:id])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+      session[:flash] = "Cannot post empty tweet"
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  # DELETE
+
+  delete "/tweets/:id/delete" do # (must validate session ID)
+    @current_user = UserHelper.current_user(session)
+    @tweet = Tweet.find(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #314, opened 2017-04-11
Done

DanL12186:


app/views/tweets/tweets.erb
```diff
@@ -0,0 +1,5 @@
+Welcome, dese r tweets.
+
+<%Tweet.all.each do |tweet|%>
```
cernanb: Convention would be to access all the tweets in the controller and save them in an instance variable

```ruby
#controller
@all_tweets = Tweet.all
```

Here in your view you would use the instance variable 

```ruby
<% @all_tweets.each do |tweet| %>
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
 require './config/environment'
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
...
+    @user = current_user #provides erb access to user and user's tweets
+    @tweets = @user.tweets
+    erb :'/tweets/tweets'
+  end
+
+  get '/tweets/new' do #Presents user with new tweet form, or sends them away if not logged in
+    logged_in? ? (erb :'/tweets/create_tweet') : (redirect '/login')
+  end
+
+  post '/tweets/new' do #creates tweet from new tweet form post request unless it's blank.
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
...
+    unless params[:content].empty?
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
...
+      @tweet = Tweet.create(user_id: current_user.id, content: params[:content]) #creates new tweet using current_user object id and params
```
```
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
...
+    if current_user.id == @tweet.user_id
+      @tweet.delete
+      redirect '/tweets'
+    else
+      redirect '/login'
+    end
+  end
+
+  helpers do
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
...
+      User.find(session[:user_id])
+    end
+
+    def logged_in? #dunno if this should be a class method
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
...
+    erb :"/tweets/show_tweet"
+  end
+
+  get "/tweets/:id" do #redirects to tweet edit URL, made necessary because I misunderstood that the spec didn't want edit/delete on the same page.
+    logged_in? ? (redirect "/tweets/#{params[:id]}/edit") : (redirect "/login") #login check/redirect login is just to satisfy spec at this juncture
+  end
+
+  get "/tweets/:id/edit" do #finds tweet and renders edit page if the user is logged in, otherwise sends packing
+    redirect '/login' unless logged_in?
+    @tweet = Tweet.find(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
...
+    erb :'/tweets/edit_tweet'
+  end
+
+  post "/tweets/:id" do #Handles POST from tweet edit form and catches no-content exception
```
cernanb: Convention for this route would be to use the `patch` http verb.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
...
+    @tweet = Tweet.find(params[:id])
+    redirect "/tweets/#{@tweet.id}/edit" if params[:content].empty? #reloads page instead of saving tweet if tweet blank
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,110 @@
...
+    @tweet.update(content: params[:content]) #otherwise modifies and saves it as expected, then redirects to /tweets
+    redirect '/tweets'
+  end
+
+  delete '/tweets/:id/delete' do #deletes tweet if if belongs to user, otherwise sends them away.
+    @tweet = Tweet.find(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #322, opened 2017-04-25
Done

latinadeveloper:


app/controllers/application_controller.rb
```diff
@@ -7,4 +7,131 @@ class ApplicationController < Sinatra::Base
     set :views, 'app/views'
   end
```

app/models/tweet.rb
```diff
@@ -0,0 +1,8 @@
+class Tweet < ActiveRecord::Base
+  validates_presence_of :content
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +7,131 @@ class ApplicationController < Sinatra::Base
...
+      redirect to '/login'
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+    def current_user
+      @current_user ||= User.find(session[:id])
```
cernanb: Great job using memoization here!

app/controllers/application_controller.rb
```diff
@@ -7,4 +7,131 @@ class ApplicationController < Sinatra::Base
...
+      @tweets = Tweet.all
+      erb :'tweets/index'
+    else
+      redirect to '/login'
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +7,131 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/new' do
+    if logged_in?
+      erb :'tweets/new'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    tweet = Tweet.new(content: params[:content], user: current_user)
```
```
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +7,131 @@ class ApplicationController < Sinatra::Base
...
+      @tweet = Tweet.find_by_id(params[:id])
+      erb :'tweets/show'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +7,131 @@ class ApplicationController < Sinatra::Base
...
+      erb :'tweets/edit'
+    else
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find_by_id(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -7,4 +7,131 @@ class ApplicationController < Sinatra::Base
...
+    end
+  end
+
+
+  get '/users/:slug' do
+    erb :'show'
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

cernanb (commented):
Hi Isis,

Regards,

Cernan

PR #324, opened 2017-04-29
Project submitted

PhreeMason:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
+class TweetController < ApplicationController
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    if logged_in?
+      @user = current_user
+      erb :'views/tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/new/tweet' do
+    if !params[:tweet][:content].blank?
```

app/controllers/application_controller.rb
```diff
@@ -4,7 +4,23 @@ class ApplicationController < Sinatra::Base
...
+    erb :'views/index'
+  end
+
+  helpers do
+    def logged_in?
+      !!current_user
+    end
+
+    def current_user
+      User.find(session[:user_id]) if session[:user_id]
```

```ruby
def current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+  get '/tweets/new' do
+    if logged_in?
+      @user = current_user
+      erb :'views/tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/new/tweet' do
```
cernanb: Typically, tweet creation would be posted to `/tweets` instead of `/tweets/new`.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    if !params[:tweet][:content].blank?
+      tweet =Tweet.create(params[:tweet])
```
cernanb: Tweet creation can be done making use of the `has_many` relationship on User:
```
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+      redirect to '/tweets'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @user = current_user
```
cernanb: It would make sense to check and make sure that the tweet belongs to the current user before taking them to the edit form here.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      @tweet = Tweet.find(params[:id])
+      erb :'views/tweets/edit_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/:id/edit' do
```
cernanb: Updating a tweet normally is a patch request to the `/tweets/:id` route.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+  get '/tweets/:id' do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
+      erb :'views/tweets/show_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/delete/:id' do
```
cernanb: Deleting a tweet would normally be a delete request to the `'/tweets/:id'` route.

PR #333, opened 2017-05-10
Project submitted

Mk-Etlinger:


app/controllers/application_controller.rb
```diff
@@ -3,8 +3,122 @@
 class ApplicationController < Sinatra::Base
 
   configure do
+    enable :sessions
     set :public_folder, 'public'
     set :views, 'app/views'
+    set :session_secret, "fwitter"
   end
```

app/models/user.rb
```diff
@@ -0,0 +1,16 @@
+class User < ActiveRecord::Base
+    has_many :tweets
+    has_secure_password
+    validates_presence_of :username, :email
```
cernanb: Great job using ActiveRecord model validations!

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,122 @@
...
+    erb :'users/index'
+  end
+
+
+  helpers do
+    def logged_in?(session)
+      !!session[:id]
+    end
+    
+    def current_user(session)
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,122 @@
...
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    erb :'users/index'
+  end
+
+
+  helpers do
+    def logged_in?(session)
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,122 @@
...
+      redirect "/login"
+    end
+  end
+
+  get '/tweets/new' do
+    redirect("/login") if !logged_in?(session)
+    erb :'tweets/new'
+  end
+
+  post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,122 @@
...
+    redirect back if params[:content].empty?
+    @tweet = Tweet.create(params)
```
```
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,122 @@
...
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
+    @tweet.delete if current_user(session).id == @tweet.user.id
+    redirect "/tweets"
+  end
+
+  get '/tweets/:id/edit' do
+    redirect("/login") if !logged_in?(session)
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,122 @@
...
+    @tweet = Tweet.find(params[:id])
+    erb :'tweets/edit'
+  end
+
+  patch '/tweets/:id/edit' do
+    redirect back if params[:content].empty?
+    @tweet = Tweet.find(params[:id])
```

PR #334, opened 2017-05-11
Done

lucketta:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+    enable :sessions
+    set :session_secret, "secret"
   end
 
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+  get '/signup' do
+    if logged_in?
+       redirect "/tweets"
+    else
+      erb :"/users/create_user"
+    end
+  end
+
+  post '/signup' do
+    if !params[:username].empty? && !params[:email].empty? &&!params[:password].empty?
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "secret"
   end
```
cernanb: You could define a current_user method here 

```ruby
def current_user
  User.find(session[:id])
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/new' do
+    if logged_in?
+      @user = User.find(session[:id])
+      erb :"/tweets/create_tweet"
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/users/:user_id' do
```
cernanb: Typically, tweet creation would be posted to `/tweets`

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      @user = User.find(params[:user_id])
+      if !params[:content].empty?
+        @user.tweets << Tweet.create(content: params[:content])
```

``` ruby
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+      @tweet = Tweet.find(params[:tweet_id])
+      erb :"/tweets/show_tweet"
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:tweet_id/edit' do
+    if logged_in?
+      @tweet = Tweet.find(params[:tweet_id])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+        redirect '/tweets'
+      end
+    end
+      redirect '/login'
+  end
+
+  patch '/tweets/:tweet_id' do
+    if logged_in?
+      @tweet = Tweet.find(params[:tweet_id])
+
```

PR #336, opened 2017-05-18
Done

nadinesk:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "password_security"
   end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+    if !session[:user_id]
+      erb :'users/create_user'
+    else
+      redirect to '/tweets'
+    end
+  end
+
+
+  post '/signup' do
+    if params[:username] == "" || params[:password] == "" || params[:email] == ""
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+    set :session_secret, "password_security"
   end
 
+  get '/' do
+    erb :index
+  end
+
+  get '/signup' do
+    
+    if !session[:user_id]
```
cernanb: Instead of inspecting the session here, it's advisable to create helper methods `current_user` and `logged_in?`.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/new' do
+    if session[:user_id]
+      erb :'tweets/create_tweet'
+    else
+      redirect "/login"
+    end
+  
+  end
+
+  post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+    if params[:content] == ""
+      redirect "tweets/new"
+    else
+      @tweet = Tweet.create(:content => params[:content])
```
```
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+      erb :'tweets/show_tweet'
+    else
+      redirect "/login"
+    end
+
+  end
+
+   get '/tweets/:id/edit' do
+    if session[:user_id]
+      @tweet = Tweet.find(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,147 @@ class ApplicationController < Sinatra::Base
...
+    end
+
+  end
+
+  patch '/tweets/:id' do      
+      @tweet =  Tweet.find_by_id(params[:id])
+      if params[:content] == ""
+        
+        redirect to("/tweets/#{@tweet.id}/edit")
+      else
```

PR #337, opened 2017-05-18
Done

andrewjford:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,75 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/login' do
+    if !logged_in?
+      erb :login
+    else
+      redirect '/tweets'
+    end
+  end
+
+  get '/logout' do
```
cernanb: Convention would be to place this action in a separate SessionsController.

app/models/tweet.rb
```diff
@@ -0,0 +1,5 @@
+class Tweet < ActiveRecord::Base
+  validates_presence_of :content
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,75 @@ class ApplicationController < Sinatra::Base
...
+      session[:user_id] = user.id
+      redirect "/tweets"
+    else
+      redirect '/login'
+    end
+  end
+
+  helpers do
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,75 @@ class ApplicationController < Sinatra::Base
...
+    end
+
+    def logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    elsif logged_in?
+      redirect "/tweets/#{params[:id]}"
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    params[:tweet][:user] = current_user
+    tweet = Tweet.new(params[:tweet])
```
cernanb: The tweet creation and user association can both be accomplished in the same step:

``` ruby
tweet = Tweet.create(content: params[:content], user: current_user)
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if !logged_in?  #had to add this in to get test to pass
+      redirect '/login' # not sure why lines 42-43 did not cover it
+    end
+    @tweet = Tweet.find(params[:id])
+    if logged_in? && @tweet.user == current_user
```
cernanb: I don't see a lot of students checking the tweet's user. Good job!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      redirect "/tweets/#{tweet.id}"
+    else
+      redirect "/tweets/new"
+    end
+  end
+
+  post '/tweets/:id' do
+    tweet = Tweet.find(params[:id])
+    if params[:tweet][:content] != ""
+      tweet.update(params[:tweet])
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      redirect "/tweets/#{tweet.id}"
+    else
+      redirect "/tweets/#{tweet.id}/edit"
+    end
+  end
+
+  post '/tweets/:id/delete' do
```
cernanb: Convention would be for this route to be a delete http request rather than a post.

PR #338, opened 2017-05-19
Project submitted

tracypholmes:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,31 @@
...
 
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+    def logged_in?
+
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,31 @@
...
+      !!session[:user_id]
+    end
+
+    def current_user
+       @current_user ||= User.find(session[:user_id])
```
cernanb: Great job using memoization!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  get '/tweets/new' do
+    if logged_in?
+      erb :'/tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params.value?('')
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+      redirect to '/tweets/new'
+    else
+      user = User.find(session[:user_id])
+      Tweet.create(user_id: user.id, content: params[:content])
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  # fin by id code 'inspired' by J
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == current_user.id
```

PR #342, opened 2017-05-25
Done

sophiav:


app/models/user.rb
```diff
@@ -0,0 +1,18 @@
+class User < ActiveRecord::Base
+  validates_presence_of :username, :email
```
cernanb: Great job using ActiveRecord model validations!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    if Helpers.logged_in?(session)
+      erb :'/tweets/new'
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do
+    @tweet = Tweet.new(content: params[:content])
+    @tweet.user_id = session[:user_id]
```
cernanb: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+      erb :'/tweets/show'
+    else
+      redirect "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do 
+    if Helpers.logged_in?(session)
+      @tweet = Tweet.find(params[:id])
+        if @tweet.user_id == session[:user_id]
```

Also, you can take advantage of the current_user method instead of using the session
`if @tweet.user_id == Helpers.current_user(session).id`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    else
+      redirect "/login"
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
+    
+    if !params[:content].empty?
+      @tweet.update(content: params[:content])
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,9 @@
+class Helpers
```
cernanb: Convention would be to define these methods inside the ApplicationController

```ruby
helpers do
  def current_user
  
  end

  def logged_in?
  
  end
end
```

PR #348, opened 2017-06-02
Done

cs-j:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,27 @@
...
 
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+    def current_user
+      @current_user ||= User.find(session[:id]) if session[:id]
```
cernanb: Great job memoizing the current user!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect "/tweets"
+    else
+      redirect "/tweets/#{tweet.id}"
+    end
+  end
+  end
+
+  post '/tweets' do
+    if logged_in?
+      @tweet = Tweet.new(content: params[:content], user_id: session[:id])
```
```
```

PR #349, opened 2017-06-02
Done Fwitter

mhornbacher:


app/controllers/login_controller.rb
```diff
@@ -0,0 +1,36 @@
+class LoginController < ApplicationController
```
cernanb: Great job splitting your actions among multiple controllers!

app/models/tweet.rb
```diff
@@ -0,0 +1,4 @@
+class Tweet < ActiveRecord::Base
+    belongs_to :user
+    validates :content, :user_id, presence: true
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,33 @@
...
+  get '/users/:user_slug' do
+    @user = User.find_by_slug(params[:user_slug])
+    erb :"users/show"
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
+    end
+    def current_user
```
cernanb: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`. You can accomplish this with Memoization.  

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,33 @@
...
+  end
+
+  get '/users/:user_slug' do
+    @user = User.find_by_slug(params[:user_slug])
+    erb :"users/show"
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:id]
```

```ruby
def logged_in?
  !!current_user
end
```

cernanb (commented):
Hi Menachem,

Regards,

Cernan

app/controllers/application_controller.rb
PCGeekBrain: I know, figured this out on my portfolio project. will definitely add it here. thanks for pointing it out 👍

PR #351, opened 2017-06-05
Done

johnfewell:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,151 @@
-require './config/environment'
-
```
cernanb: Great job splitting your actions among multiple controllers!

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,151 @@
...
+    get '/tweets/new' do
+      if session[:user_id]
+        erb :'tweets/create_tweet'
+      else
+        redirect to '/login'
+      end
+    end
+
+    post '/tweets' do
+      if params[:content] == ""
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,151 @@
...
+          redirect to '/tweets'
+        end
+      else
+        redirect to '/login'
+      end
+    end
+
+
+    helpers do
+       def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,151 @@
...
+      else
+        redirect to '/login'
+      end
+    end
+
+    post '/tweets' do
+      if params[:content] == ""
+        redirect to "/tweets/new"
+      else
+        @tweet = Tweet.create(:content => params[:content], :user_id => current_user.id)
```
```
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,151 @@
...
+        erb :'tweets/show_tweet'
+      else
+        redirect to '/login'
+      end
+    end
+
+    get '/tweets/:id/edit' do
+      if session[:user_id]
+        @tweet = Tweet.find(params[:id])
+        if @tweet.user_id == session[:user_id]
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,151 @@
...
+    end
+
+    get '/users/:slug' do
+      @user = User.find_by_slug(params[:slug])
+      erb :'users/show'
+    end
+
+    delete '/tweets/:id/delete' do
+      if session[:user_id]
+        @tweet = Tweet.find(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #353, opened 2017-06-07
Done

cquinones100:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,121 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "secret"
   end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,121 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/new' do
+    if current_user
+      erb :'/tweets/new'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    if params[:content].empty?
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,121 @@ class ApplicationController < Sinatra::Base
...
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
+    if current_user == @tweet.user
+      @tweet.destroy
+    end
+    redirect to '/tweets'
+  end
+
+  def current_user
+    @user ||= User.find(session[:id]) unless session[:id].nil?
```
cernanb: Great job memoizing current user!

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,121 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/tweets/new' do
+    if current_user
+      erb :'/tweets/new'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,121 @@ class ApplicationController < Sinatra::Base
...
+    if params[:content].empty?
+      redirect to "/tweets/new"
+    else
+      @tweet = Tweet.create(params)
+      if current_user
+        @tweet.user = current_user
```

``` ruby
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,121 @@ class ApplicationController < Sinatra::Base
...
+  get '/tweets/:id/edit' do
+    if current_user
+      @tweet = Tweet.find(params[:id])
+      erb :'tweets/edit'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/:id/edit' do
```
cernanb: Updating a tweet would normally be a patch request to '/tweets/:id'

```ruby
patch '/tweets/:id/' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,121 @@ class ApplicationController < Sinatra::Base
...
+    @tweet = Tweet.find(params[:id])
+    @tweet.update(content: params[:content])
+    erb :'tweets/show'
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
+    if current_user == @tweet.user
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #355, opened 2017-06-09
Done

asaki444:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,62 @@
+class TweetController < ApplicationController
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+     tweet = Tweet.find_by(user_id: session[:id])
+     tweet.content = params['content']
+     tweet.save
+   else
+     redirect "tweets/#{session[:id]}/edit"
+   end
+  end
+
+  post '/tweets/' do
+    if params['content'] != ""
```

PR #391, opened 2017-07-20
Project submitted

repromancer:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,102 @@
+class TweetsController < ApplicationController
+
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+  get '/tweets/new' do
+    if_logged_in do
+      erb :'/tweets/new'
+    end
+  end
+
+  post '/tweets' do
+    current_user.tap do |user|
+
+      if params[:content].chars.any?
```

app/controllers/helpers/login_utils.rb
```diff
@@ -0,0 +1,39 @@
+module LoginUtils
+
+  def current_user
+    User.find(session[:user_id])
```
cernanb: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/controllers/helpers/login_utils.rb
```diff
@@ -0,0 +1,39 @@
...
+  def if_logged_in
+    if session[:user_id]
+      yield
+    else
+      flash[:message] = "Please log in."
+      redirect '/login'
+    end
+  end
+
+  def logged_in?
```
cernanb: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+  end
+
+
+  get '/tweets/:id/edit' do
+    if_logged_in do
+
+      @tweet = Tweet.find(params[:id])
+
+      if current_user.owns?(@tweet)
```
cernanb: I don't see a lot of students checking the tweet's ownership. Good job!

app/controllers/tweets_controller.rb
repromancer: Whoa, thanks for mentioning that! I had basic validations in the User class, but I didn't realize you could do so much more with them!

PR #399, opened 2017-08-02
Done

infinitelooplabsLLC:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
+require 'rack-flash'
+
```

app/models/tweet.rb
```diff
@@ -0,0 +1,6 @@
+class Tweet < ActiveRecord::Base
+
+  belongs_to :user
+  validates :content, presence: true
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+  get '/tweets' do
+    redirect to '/login' unless logged_in?
+
+    @tweets = Tweet.all
+    erb :'/tweets/tweets'
+  end
+
+  post '/tweets' do
+    @tweet = Tweet.new(content: params[:content], user_id: current_user.id)
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    @tweet = Tweet.find(params[:id])
+    erb :'/tweets/show_tweet'
+  end
+
+  get '/tweets/:id/edit' do
+    redirect to '/login' unless logged_in?
+
+    @tweet = Tweet.find(params[:id])
+
+    if @tweet.user.id != current_user.id
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      flash[:notice] = "Invalid content. Please try again."
+      redirect to "/tweets/#{@tweet.id}/edit"
+    elsif @tweet.update(content: params[:content])
+      redirect to "/tweets/#{@tweet.id}"
+    else
+      redirect to '/tweets'
+    end
+  end
+
+  delete '/tweets/:id/delete' do
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

cernanb (commented):
Hi Zain,

Regards,

Cernan

PR #403, opened 2017-08-04
Done

byung90:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
+class TweetsController < ApplicationController
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    if logged_in?
+      @tweet = Tweet.find_by(params[:id])
+      erb :'tweets/edit_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] == ""
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      erb :'tweets/show_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by(params[:id])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = params[:content]
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #412, opened 2017-08-14
Project submitted

soheecho94:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect to "/login"
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] == ""
+      redirect to "/tweets/new"
+    else
+      @tweet = Tweet.new(params)
+      @tweet.user_id = session[:user_id]
```
cernanb: The tweet creation and user association can both be accomplished in the same step:

``` ruby

File: app/controllers/tweet_controller.rb [hunk 2]
```diff
@@ -0,0 +1,80 @@
...
+      erb :"tweets/show_tweet"
+    else
+      redirect to "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user_id == current_user.id
```
cernanb: I don't see a lot of students checking the tweet's user id. Good job!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect to "/login"
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find_by_id(params[:id])
+    if params[:content].empty?
+      redirect to "/tweets/#{@tweet.id}/edit"
+    else
+      @tweet.update(content: params[:content])
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect to "/tweets/#{@tweet.id}"
+    end
+  end
+
+  delete '/tweets/:id/delete'do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
```
cernanb: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #414, opened 2017-08-16
Project submitted

andori8:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  patch '/tweets/:id/edit' do
+    @tweet = Tweet.find(params[:id])
+    @tweet.update(content: params[:content])
+    @tweet.save
+    redirect "/tweets"
+  end
+
+  post '/tweets/new' do
```
cernanb: Typically, tweet creation would be posted to `/tweets` instead of `/tweets/new`.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    if params[:content].empty?
+      redirect "/tweets/new"
+    else
+      tweet = Tweet.create(content: params[:content], user_id: session[:user_id])
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      @tweet = Tweet.find(params[:id])
+      erb :"/tweets/show"
+    else
+      redirect "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      erb :"/tweets/edit"
+    else
+      redirect "/login"
+    end
+  end
+
+  patch '/tweets/:id/edit' do
+    @tweet = Tweet.find(params[:id])
+    @tweet.update(content: params[:content])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      redirect "/tweets/new"
+    else
+      tweet = Tweet.create(content: params[:content], user_id: session[:user_id])
+      redirect "/tweets"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
+    @tweet.delete if @tweet.user_id == session[:user_id]
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #521, opened 2017-11-13
hopefully this will work!

ddeleon267:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,76 @@
+class TweetController < ApplicationController
+
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+  get '/tweets/new' do
+    if logged_in?(session)
+      erb :'tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] != ""
```

app/models/user.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def slug
+    username.downcase.gsub(" ", "-")
+  end
+
+  def self.find_by_slug(slug)
+    self.all.find{|user| user.slug == slug}
+  end
+
+  def self.current_user(session)
```
cernanb: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+      tweet = Tweet.new(params)
+      tweet.user_id = session[:id]
```
cernanb: Another way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+      erb :'/tweets/show_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  ############ UPDATE ###########
+  get '/tweets/:id/edit' do
+    @tweet = Tweet.find_by(id: params[:id])
+    if logged_in?(session) && User.current_user(session).id == @tweet.user_id
```
cernanb: I don't see a lot of students checking the tweet's user id. Good job!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+      erb :'/tweets/edit_tweet'
+    elsif logged_in?(session) && User.current_user(session).id != @tweet.user_id
+      redirect to '/tweets'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/:id' do
```
cernanb: Typically an update would be a patch request rather than a post request.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+    @tweet = Tweet.find(params[:id])
+    if params[:content] != ""
+      @tweet.content = params[:content]
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+      @tweet.save
+      redirect to "/tweets/#{@tweet.id}"
+    else
+      redirect to "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  ############ DELETE ###########
+  post '/tweets/:id/delete' do
```
cernanb: Typically deleting a tweet would be a delete request rather than post.

PR #526, opened 2017-11-17
Project submitted

GusVilla303:


app/controllers/application_controller.rb
```diff
@@ -3,8 +3,156 @@
 class ApplicationController < Sinatra::Base
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,156 @@
...
+    if logged_in?
+      session.clear
+      redirect '/login'
+    else
+      redirect '/'
+    end
+  end
+
+  post '/tweets' do
+    if !params[:content].blank?
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,156 @@
...
+      end
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,156 @@
...
+        redirect '/tweets'
+      else
+        redirect "/tweets/#{tweet.id}"
+      end
+    end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,156 @@
...
+      redirect '/login'
+    else
+      redirect '/'
+    end
+  end
+
+  post '/tweets' do
+    if !params[:content].blank?
+      @tweet = Tweet.create(content: params[:content])
+      current_user.tweets << @tweet
```
```
```
Both approaches are good, but in some cases you may find this option useful

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,156 @@
...
+
+      redirect "/tweets/#{@tweet.id}"
+    else
+      redirect '/tweets/new'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,156 @@
...
+
+      erb :'tweets/edit_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/:id' do
```
cernanb: Typically updating a tweet would be a patch http request rather than post.

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,156 @@
...
+    @tweet = Tweet.find(params[:id])
+    if !params[:content].blank?
+      @tweet.update(content: params[:content])
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,156 @@
...
+
+      redirect "tweets/#{@tweet.id}"
+    else
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  post '/tweets/:id/delete' do
```
cernanb: For deleting a tweet convention would be for this to be a delete http request.

PR #543, opened 2017-12-01
Done

omikolaj:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,31 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "password_security"
   end
```
cernanb: Great job splitting your actions among multiple controllers!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+        if logged_in? && !params["content"].empty?
+            @tweet.update(:content => params["content"])
+            redirect "/tweets/#{@tweet.id}/edit"
+        else
+            redirect "/tweets/#{@tweet.id}/edit"
+        end        
+    end
+
+    post '/tweets' do
+        if logged_in? && !params["content"].empty?
```
cernanb: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks in your model more easily.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,31 @@ class ApplicationController < Sinatra::Base
...
 
+
+  helpers do
+
+    def current_user
+      @current_user ||= User.find_by(id: session[:user_id])
```
cernanb: Excellent work using memoization!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+            @tweet = Tweet.find_by_id(params["id"])
+            erb :"tweets/show_tweet"
+        else
+            redirect "/login"
+        end
+    end
+    
+    get '/tweets/:id/edit' do
+        if logged_in?
+            @tweet = Tweet.find_by_id(params["id"])
```
cernanb: It's good practice to verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+            erb :'tweets/edit_tweet'
+        else
+            redirect '/login'
+        end
+    end
+
+    patch '/tweets/:id' do
+        @tweet = Tweet.find_by_id(params["id"])
```

omikolaj:
Thanks for reviewing my Fwitter project. I appreciate the code review. Mostly all of your suggestions I have implemented in my Portfolio Project already, with the exception of your comment "There are other ways to submit a PATCH/POST request than going through an HTML form...". I will have to research this.

PR #683, opened 2018-03-19
Done

anthonygharvey:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,42 @@
...
+  #==================== INDEX =============================
+  get '/' do
+    erb :'index'
+  end
+  #--------------------------------------------------------
+
+ 
+  #==================== HELPERS ===========================
+  helpers do
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,42 @@
...
+      User.find(session[:user_id])
+    end
+  
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+			redirect to "/login"
+		end
+	end
+	#--------------------------------------------------------
+
+
+	#==================== EDIT ==============================
+	get '/tweets/:tweetid/edit' do
+		if logged_in?
+			@tweet = Tweet.find(params[:tweetid])
```
kkterai: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+			erb :'/tweets/edit_tweet'
+		else
+			redirect to "/login"
+		end
+	end
+
+	patch '/tweets/:tweetid' do
+		if !params[:content].empty?
+			@tweet = Tweet.find(params[:tweetid])
```
kkterai: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+			redirect to "/tweets/#{params[:tweetid]}/edit"
+		end
+	end
+	#--------------------------------------------------------
+
+
+	#==================== DELETE ============================
+	delete '/tweets/:tweetid' do
+		if logged_in?
+			current_user.tweets.find_by(params[:tweetid]).delete
```
kkterai: Nice job finding tweets that are already only associated to the `current_user`!

app/views/layout.erb
```diff
@@ -2,14 +2,40 @@
 <html>
   <head>
     <title>Fwitter</title>
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootswatch/4.0.0/lumen/bootstrap.min.css">
```
kkterai: Above and beyond!

PR #685, opened 2018-03-21
Done

cdhenry:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
+class TweetsController < ApplicationController
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  get '/tweets' do
+    if session[:user_id]
+      @user = User.find_by_id(session[:user_id])
```
cernanb: Instead of referencing the session object you can build a `current_user` method and place that in a helpers block in your application controller. 

```ruby
def current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  get '/tweets/new' do
+    if session[:user_id]
+      erb :'tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/new' do
```
cernanb: Typically, tweet creation would be posted to `/tweets` instead of `/tweets/new`.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    if user = User.find_by_id(session[:user_id])
+      if user.tweets << Tweet.new(content: params[:content])
```
cernanb: The tweet creation and user association can both be accomplished in the same step:

``` ruby
@tweet = Tweet.create(content: params[:content], user_id: user.id)
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if session[:user_id]
+      user = User.find_by_id(session[:user_id])
+      @tweet = Tweet.find_by_id(params[:id])
+      if user.tweets.include?(@tweet)
```
cernanb: Another way to check for this is to look at the user_id value for the tweet. 

`if user.id == @tweet.user.id`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    else
+      redirect to '/login'
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    if session[:user_id]
+      user = User.find_by_id(session[:user_id])
+      tweet = Tweet.find_by_id(params[:id])
+      if i = user.tweets.find_index(tweet)
```

```ruby
tweet = user.tweets.find_by_id(params[:id])
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

cdhenry:
Thanks Cernan, this is super helpful!

Quick question, was I supposed to get one of these emails from an
instructor for the secondary CLI OO Ruby Project as well?  I think it was
called TicTacToe with AI, but I can't remember at the moment.

Just want to make sure I'm not missing anything.

Best,
Chris

PR #686, opened 2018-03-21
Done

jendobes:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,34 @@ class ApplicationController < Sinatra::Base
...
   end
 
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+      def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,34 @@ class ApplicationController < Sinatra::Base
...
+        !!session[:user_id]
+      end
+
+      def current_user
```

```ruby
def current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    if logged_in?
+      erb :'/tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if !params['content'].empty?
+      @tweet = Tweet.create(:content => params['content'], :user_id => current_user.id)
```
kkterai: Another way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  get '/tweets/:id' do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
+      erb :'/tweets/show_tweets'
+    else
+      redirect to "/login"
+    end
+  end
+
+  get '/tweets/:id/edit' do
```
kkterai: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      erb :'/tweets/edit_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/:id' do
+      if params['content'] == ""
+        redirect to "/tweets/#{params[:id]}/edit"
+      else
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+        @tweet.update(:content => params['content'])
+        @tweet.save
+        redirect to "/tweets/#{@tweet.id}"
+      end
+  end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet && @tweet.user == current_user
```
kkterai: Good job ensuring that the tweet to be deleted belongs to the current_user!

kkterai:
Hi Jennifer,

Great job on your Fwitter Project! I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Feel free to go back through your code and update it using my comments. This is not required. If you have any questions you can contact me by email: kenlyn@flatironschool.com or on slack: @kkterai.

Regards,

Kenlyn Terai

PR #687, opened 2018-03-21
Done

unenlightened:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
+class TweetsController < ApplicationController
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      else
+        redirect "/tweets/#{@tweet.id}"
+      end
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do
+    redirect "/tweets/new" if params[:content].empty?
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,26 @@ class ApplicationController < Sinatra::Base
...
+      erb :index
+    end
+  end
+
+  helpers do
+		def logged_in?
+			!!session[:user_id]
+		end
+
+		def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,26 @@ class ApplicationController < Sinatra::Base
...
+    if logged_in?
+      redirect "/tweets"
+    else
+      erb :index
+    end
+  end
+
+  helpers do
+		def logged_in?
+			!!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    @tweet = Tweet.create(content: params[:content], user_id: current_user.id)
```
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  post '/tweets/:id' do
+    redirect "/tweets/#{params[:id]}/edit" if params[:content].empty?
+
+    @tweet = Tweet.find(params[:id])
+    @tweet.update(content: params[:content])
+    redirect "/tweets/#{@tweet.id}"
+  end
+
+  post '/tweets/:id/delete' do
```
cernanb: Convention for this would be a `delete` to `'/tweets/:id'`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  post '/tweets' do
+    redirect "/tweets/new" if params[:content].empty?
+
+    @tweet = Tweet.create(content: params[:content], user_id: current_user.id)
+    redirect "/tweets/#{@tweet.id}"
+  end
+
+  post '/tweets/:id' do
```
cernanb: Convention for this would be a `patch`

app/controllers/tweets_controller.rb
unenlightened: Thank you for the great review. Just made those changes. Good to learn about ActiveRecord Validations and shortcuts for creating objects.

PR #690, opened 2018-03-24
Done

safuya:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,44 @@
+class TweetController < ApplicationController
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    erb :'tweets/index'
+  end
+
+  get '/tweets/new' do
+    redirect '/login' unless session.key?(:user_id)
+    erb :'tweets/new'
+  end
+
+  post '/tweets' do
+    redirect '/login' unless session.key?(:user_id)
```
weezwo: A `logged_in?` method defined as a helper in the ApplicationController could be useful here, which is preferable to referring to the session object directly.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  get '/tweets/:id' do
+    redirect '/login' unless session.key?(:user_id)
+    @tweet = Tweet.find(params[:id])
+    erb :'tweets/show'
+  end
+
+  get '/tweets/:id/edit' do
+    redirect '/login' unless session.key?(:user_id)
```
weezwo: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    @tweet = Tweet.find(params[:id])
+    erb :'tweets/edit'
+  end
+
+  patch '/tweets/:id' do
```
weezwo: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    redirect "/tweets/#{params[:id]}/edit" if params[:content].empty?
+    tweet = Tweet.find(params[:id])
+    tweet.update(content: params[:content])
+  end
+
+  delete '/tweets/:id/delete' do
+    redirect '/tweets' unless session.key?(:user_id)
```
weezwo: Shouldn't you verify a tweet belongs to the current_user (and not just that any user is logged in) before allowing them to delete?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  get '/tweets/new' do
+    redirect '/login' unless session.key?(:user_id)
+    erb :'tweets/new'
+  end
+
+  post '/tweets' do
+    redirect '/login' unless session.key?(:user_id)
+    redirect '/tweets/new' if params[:content].empty?
+    Tweet.create(content: params[:content], user_id: session[:user_id])
```
weezwo: A `current_user` helper method could be useful here.

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    erb :'users/index'
+  end
+
+  get '/signup' do
+    redirect '/tweets' if session.key?(:user_id)
+    erb :'users/new'
+  end
+
+  post '/signup' do
+    redirect '/signup' if params[:username].empty?
```
weezwo: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,9 @@
...
-
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
   end
-
-end
\ No newline at end of file
+end
```
weezwo: Here's where you would define your helpers -- I think `current_user` and `logged_in?` would be particularly helpful.

weezwo (commented):
Nice job!! Left a few comments for you. There's no need to submit this project again -- these are just a few tips for the future.

PR #691, opened 2018-03-25
Done

AlphaDaniel:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,57 @@
...
+    
+    erb :"index"
+  end
+#==============================helpers================================= 
+  def logged_in? 
+    !!current_user
+  end
+    
+  def current_user 
+    @current_user ||= User.find(session[:user_id]) if session[:user_id]
```
weezwo: Nice job memoizing `current_user`!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  get '/tweets/new' do 
+    redirect '/login' if !logged_in?
+    
+    erb :'tweets/create_tweet'
+  end
+#-------------------------------------------------------------------- 
+  post '/tweets' do 
+    redirect '/tweets/new' if empty?(params)
+    
+    Tweet.create(content: params[:content], user_id: current_user.id)
```
weezwo: You can also specify `user` directly here -- for example:

```
 Tweet.create(content: params[:content], user: current_user)
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    redirect "/tweets/#{@tweet.id}/edit" if params[:content].empty?
+    
+    @tweet.update(content: params[:content])
+    redirect "/tweets/#{@tweet.id}"
+  end
+#=============================delete================================= 
+  delete '/tweets/:id/delete' do 
+    @tweet = tweet(params[:id])
+    
+    @tweet.destroy! if @tweet.user_id == current_user.id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    
+    erb :'/tweets/show_tweet'
+  end
+#==============================edit================================== 
+  get '/tweets/:id/edit' do 
+    redirect '/login' if !logged_in? 
+    
+    @tweet = tweet(params[:id])
+    
+    redirect '/tweets' if @tweet.user_id != current_user.id
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  get '/tweets/:id/edit' do 
+    redirect '/login' if !logged_in? 
+    
+    @tweet = tweet(params[:id])
+    
+    redirect '/tweets' if @tweet.user_id != current_user.id  
+    erb :'/tweets/edit_tweet'
+  end
+#=============================update================================= 
+  patch '/tweets/:id' do
```

app/views/layout.erb
```diff
@@ -2,14 +2,29 @@
 <html>
   <head>
     <title>Fwitter</title>
+    <link href="/css/main.css" rel="stylesheet">
```
weezwo: Above and beyond!!

app/views/layout.erb
```diff
@@ -2,14 +2,29 @@
...
   </head>
   <body>
+    <nav>
+      <a href="/">Home</a>
+      <% if session[:user_id].present? %>
```
weezwo: You can use your `logged_in?` helper method here, which is preferable to referring to the session object directly.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,57 @@
+class TweetsController < ApplicationController
```
weezwo: Great job splitting your actions among multiple controllers!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+#===============================new================================== 
+  get '/tweets/new' do 
+    redirect '/login' if !logged_in?
+    
+    erb :'tweets/create_tweet'
+  end
+#-------------------------------------------------------------------- 
+  post '/tweets' do 
+    redirect '/tweets/new' if empty?(params)
```

app/controllers/tweets_controller.rb
AlphaDaniel: ah nice!! and yep makes sense

app/controllers/tweets_controller.rb
AlphaDaniel: that's an awesome tip especially since I default to using find() when only need the id as it's prettier. Will definitely keep this in mind :).

app/controllers/tweets_controller.rb
AlphaDaniel: :) thanks

app/views/layout.erb
AlphaDaniel: Yea not going to lie while styling.. I coded wam bam as it wasn't a focus of the lab and just wanted to have a little fun but I'm sure I wouldv'e if I was really throwing all focus on it see below from my sinatra project where I'm using it exactly like that :)

https://github.com/AlphaDaniel/snalgos/blob/a7bdc78edceb11f62a7dff6249fd49c8617e4083/app/views/layout.erb#L37

app/controllers/tweets_controller.rb
AlphaDaniel: thanks and yea I learned about protecting these routes also after I finished this lab when doing some light reading on url hacking all my future patch/post requests will keep this in mind .. thanks :)

app/controllers/tweets_controller.rb
AlphaDaniel: Also learned about this after fwitter lol using it here 

https://github.com/AlphaDaniel/snalgos/blob/a7bdc78edceb11f62a7dff6249fd49c8617e4083/app/models/user.rb#L7

PR #697, opened 2018-03-30
Done

alyssaswatson:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,23 @@ class ApplicationController < Sinatra::Base
...
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+    def logged_in?
+      !!current_user
+    end
+
+    def current_user
```
kkterai: Good job memoizing the current_user!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,93 @@
+require './config/environment'
+
+class TweetController < ApplicationController
+
+  get '/tweets' do
+    if session[:user_id]
```
kkterai: You can use `if current_user` here. Because in the lifespan of a request, you might call `session[:user_id]` multiple times, we'd prefer to only hit the database once via a call to `current_user` 

This pattern of conditionally setting an instance variable or returning it if it's already defined using the `||=` (or-equals) operator as you did in your `current_user` method is called Memoization ([read more](http://gavinmiller.io/2013/basics-of-ruby-memoization/) and [here](http://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/)).

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+      erb :show
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    #binding.pry
+    if logged_in?
+      @tweet = Tweet.find_by(params[:id])
```
kkterai: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+  end
+
+  patch '/tweets/:id' do
+    if logged_in?
+      if params[:content] == ""
+        redirect to "/tweets/#{params[:id]}/edit"
+      else
+        @tweet = Tweet.find_by_id(params[:id])
+        if @tweet && @tweet.user == current_user
```
kkterai: I don't see a lot of students checking the tweet's user id here. Good job!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+      end
+    else
+      redirect to '/login'
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet && @tweet.user == current_user
```
kkterai: Same comment as patch `tweets/:id`: I don't see a lot of students checking the tweet's user id. Good job!

PR #701, opened 2018-03-31
Done

heislercreative:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
+class TweetsController < ApplicationController
```
weezwo: Great job splitting your actions among multiple controllers!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+  get '/tweets/new' do
+    if !logged_in?
+      redirect to '/login'
+    else
+      erb :'/tweets/new'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] == ""
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      redirect to 'tweets/new'
+    else
+      @tweet = current_user.tweets.build(params)
```
weezwo: Another way to do this is to make use of the `.create` method provided by the `has_many` relationship on User:
```
```

That way you don't have to call `.save` on the `current_user`. :)

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+  get '/tweets/:id' do
+    if !logged_in?
+      redirect to '/login'
+    else
+      @tweet = Tweet.find_by_id(params[:id])
+      erb :'/tweets/show'
+    end
+  end
+
+  get '/tweets/:id/edit' do
```
weezwo: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    else
+      if current_tweet
+        erb :'/tweets/edit'
+      else
+        redirect to '/tweets'
+      end
+    end
+  end
+
+  patch '/tweets/:id' do
```

app/views/layout.erb
```diff
@@ -2,14 +2,30 @@
 <html>
   <head>
     <title>Fwitter</title>
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.2.0/css/bootstrap.min.css">
```

PR #702, opened 2018-03-31
Done

amandarebecca89:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
 require './config/environment'
-
+require 'pry'
 class ApplicationController < Sinatra::Base
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,10 @@
+class Helpers
```
weezwo: Sinatra helpers are typically defined in a `helpers` block in the application controller. That way you won't have to pass the session -- they will already have access to it. :)

app/helpers/helpers.rb
```diff
@@ -0,0 +1,10 @@
...
+    def self.current_user(session)
+        @user = User.find_by(id: session[:id])
```
weezwo: Because in the lifespan of a request, you might call `current_user` multiple times, we'd prefer this method to only hit the database once and then 'cache' the user instance into an instance variable that could be returned on any subsequent call to `current_user`.

```ruby
def current_user
end
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,10 @@
...
+        @user
+    end
+
+    def self.is_logged_in?(session)
+        !!User.find_by(id: session[:id])
```
weezwo: It's better to rely on the successful call to `current_user` than only introspecting on the `session[:id]` value. In theory, a user might have an invalid `id` in their session which would make `logged_in?` return true even though that user would not be findable later. Also, because `current_user` is memoized, it makes sense to call it as early in the request cycle as possible, which generally is the first call `logged_in?`.

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+  get '/login' do
+    if Helpers.is_logged_in?(session)
+      redirect '/tweets'
+    else
+      erb :login
+    end
+  end
+
+  post '/signup' do
+    if !params['username'].empty? && !params['email'].empty? && !params['password'].empty?
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+      erb :'/tweets/show'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    @user = Helpers.current_user(session)
+    @tweet = Tweet.find_by(id=params[:id])
+    if Helpers.is_logged_in?(session) && @tweet.user_id == @user.id
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+      erb :'/tweets/edit'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets/:id/edit' do
```
weezwo: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    erb :'users/show'
+  end
+
+  post '/tweets/:id/delete' do
+    @tweet = Tweet.find_by(id: params[:id])
+    @tweet.delete if @tweet.user == Helpers.current_user(session)
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #704, opened 2018-04-02
Done

yanchanllin:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+  end
+
+  post '/tweets' do
+    user = current_user
+    if params["content"].empty?
+      flash[:empty_tweet] = "Please enter content for your tweet"
+      redirect to '/tweets/new'
+    end
+    tweet = Tweet.create(:content => params["content"], :user_id => user.id)
+
```
kkterai: Another way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+    erb :"tweets/show_tweet"
+  end
+
+  get '/tweets/:id/edit' do
+    if !is_logged_in?
+      redirect to '/login'
+    end
+    @tweet = Tweet.find(params[:id])
+    if current_user.id != @tweet.user_id
+      flash[:wrong_user_edit] = "Sorry you can only edit your own tweets"
```
kkterai: I don't see a lot of students checking the tweet's user id. Good job!

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,160 @@
...
+      redirect to '/tweets'
+    end
+    erb :"tweets/edit_tweet"
+  end
+
+  patch '/tweets/:id' do
+    tweet = Tweet.find(params[:id])
```
kkterai: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

PR #707, opened 2018-04-02
Done

sakhtar1:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,90 @@
+class TweetsController < ApplicationController
```
kkterai: Great job splitting your actions among multiple controllers!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+      redirect '/login'
+    end
+  end
+
+  post'/tweets' do
+    if logged_in?
+      if params[:content].empty?
+        redirect "/tweets/new"
+      else
+        @user = User.find_by(id: session[:user_id])
```
kkterai: Prefer to use the memoized value in `current_user` over looking at `session[:user_id]`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+        @tweet = Tweet.create(content: params[:content], user_id: @user.id)
```
kkterai: Another way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet && @tweet.user == current_user
+        erb :'tweets/edit_tweet'
```
kkterai: I don't see a lot of students checking the tweet's user id. Good job!

PR #708, opened 2018-04-02
Done

mietone:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,96 @@
+class TweetsController < ApplicationController
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+  get '/tweets/new' do
+    if logged_in?
+      erb :'/tweets/create_tweet'
+    else
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do
+    @tweet = current_user.tweets.build(content: params[:content])
```
cernanb: Great job using the has_many association to build a new tweet!

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,31 @@ class ApplicationController < Sinatra::Base
...
+
+
+
+  helpers do
+
+    def logged_in?
+      !!current_user
+    end
+
+    def current_user
```
cernanb: Great work using memoization!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+      redirect "/login"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by(id: params[:id])
+
+    if logged_in?
+      # binding.pry
+      if @tweet && @tweet.user == current_user
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #710, opened 2018-04-03
Done

rhhaynes:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,62 @@
+class TweetController < ApplicationController
```
kkterai: Great job splitting your actions among multiple controllers!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+      if !!@tweet && current_user.id == @tweet.user.id then erb :"tweet/edit"
+      else redirect "/tweets"
+      end
+    end
+  end
+  
+  patch "/tweets/:id" do
+    tweet = Tweet.find_by(:id => params["id"])
+    if !!tweet && current_user.id == tweet.user.id
+      if params["content"].strip.empty?
```
kkterai: I don't see a lot of students checking the tweet's user id here. Good job!

app/models/user.rb
```diff
@@ -0,0 +1,13 @@
+class User < ActiveRecord::Base
+  validates_presence_of :username, :email, :password
```
kkterai: Great job using ActiveRecord model validations!

PR #711, opened 2018-04-03
Done

johnwojo:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if logged_in?
+      if params[:content] == ""
+        redirect to '/tweets/new'
+      else
+        @tweet = current_user.tweets.build(content: params[:content])
```
kkterai: Good job making use of the `has_many` relationship on User! And I don't often see students using `build` at this point in the curriculum.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      redirect to '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    if logged_in?
+      if params[:content] == ""
+        redirect to "/tweets/#{params[:id]}/edit"
+      else
+        @tweet = Tweet.find_by_id(params[:id])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+    else
+      redirect to '/login'
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet && @tweet.user == current_user
```
kkterai: Good job checking that the tweet to be deleted is the current_user's!

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,54 @@
+require './config/environment'
+class UsersController < ApplicationController
```
kkterai: Great job splitting your actions among multiple controllers!

PR #715, opened 2018-04-06
Done

KimGonzales:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,26 @@
 require './config/environment'
+require 'rack-flash'
 
 class ApplicationController < Sinatra::Base
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  # User Registration ------------------------------------------------------- 
+
+  get '/signup' do
+    redirect to '/tweets' if logged_in?
+    erb :'users/create_user'
+  end 
+
+  post '/signup' do 
+    if !params[:username].empty? && !params[:email].empty? && !params[:password].empty?
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    logged_in? ? (erb :'tweets/create_tweet') : (redirect to '/login')
+  end 
+
+  post '/tweets' do
+    if params[:content].empty?
+      flash[:message] ="Provide some content to post a tweet!"
+      redirect to '/tweets/new'
+    elsif logged_in?
+      @user = current_user
+      @tweet =Tweet.create(params)
```
weezwo: Another way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    @tweet = Tweet.find_by_id(params[:id])
+    erb :'tweets/show_tweet'
+  end 
+
+  # Update --------------------------------------------------------------------
+
+  get '/tweets/:id/edit' do 
+    redirect to '/login' if !logged_in?
+    @tweet = Tweet.find(params[:id])
+    if logged_in? && @tweet.user != current_user
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      flash[:notice] = "You cannot change another user's content, silly"
+      redirect to "/tweets/#{@tweet.id}"
+    end
+    erb :'tweets/edit_tweet'
+  end 
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    redirect to "/tweets/#{@tweet.id}/edit" if params[:content].empty?
+    @tweet.update(content: params[:content])
+    redirect to "/tweets/#{@tweet.id}"
+  end 
+
+  # Delete --------------------------------------------------------------------
+
+  post '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
+    if logged_in? && @tweet.user == current_user
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/views/layout.erb
```diff
@@ -2,13 +2,17 @@
 <html>
   <head>
     <title>Fwitter</title>
+    <link rel="stylesheet" type="text/css" href="/css/styles.css">
```

PR #716, opened 2018-04-07
Done

jokerdelta:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,28 @@
...
 
-end
\ No newline at end of file
+    get '/' do
+      erb :'index'
+    end
+
+    	helpers do
+    		def logged_in?
+    			!!session[:user_id]
```
kkterai: You can use your `logged_in?` helper method here, which, when it calls a memoized value of current_user, is preferable to referring to the session object directly.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,28 @@
...
+    		end
+
+    		def current_user
```

```ruby
def current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,86 @@
+class TweetController < ApplicationController
```
kkterai: Great job splitting your actions among multiple controllers!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+
+
+  get '/tweets' do
+    if logged_in?
+      @user = User.find_by(id: session[:user_id])
```
kkterai: Can do `@user = current_user` here

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+      erb :'tweets/create_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets' do
+    if logged_in?
+      if params[:content] == " " #does not allow creation of new tweet if contents are blank
```
kkterai: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+        redirect to '/tweets/new'
+      else
+        @tweet = Tweet.create(content: params[:content], user_id: session[:user_id])
```
kkterai: Another way to do this is to make use of the `has_many` relationship on User:
```
```

PR #719, opened 2018-04-09
Project submitted

rhogotargaryen:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,70 @@
+class TweetController < ApplicationController
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    if logged_in?
+      current_user
+      erb :'tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+  
+  post '/tweets' do
+    if params[:content] != ""
```
weezwo: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      current_user
+      erb :'tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+  
+  post '/tweets' do
+    if params[:content] != "" 
+      Tweet.create(content: params[:content], user_id: current_user.id)
```
weezwo: Another way to do this is to make use of the `has_many` relationship on User:
```
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    if logged_in?
+      @tweet = Tweet.find_by(id: params[:tweet_id])
+    else
+      redirect '/login'
+    end
+    erb :'tweets/show_tweet'
+  end
+  
+  get '/tweets/:tweet_id/edit' do
+    if logged_in?
```
weezwo: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      redirect '/login'
+    end
+    erb :'/tweets/edit_tweet'
+  end
+  
+  post '/tweets/:tweet_id' do
+    if logged_in?
+      redirect "tweets/#{params[:tweet_id]}/edit" if params[:content] == ""
+      @tweet = Tweet.find_by_id(params[:tweet_id])
+      @tweet.content = params[:content]
```
weezwo: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    else
+      redirect '/login'
+    end
+    erb :'tweets/show_tweet'
+  end
+  
+  post '/tweets/:tweet_id/delete' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:tweet_id])
+      @tweet.delete if @current_user.id == @tweet.user.id
```
weezwo: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/views/layout.erb
```diff
@@ -1,15 +1,21 @@
...
-  </body>
+
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+
+  <title>The Flatiron Twitter "Fwitter"</title>
+  <meta name="Flatiorn school's pre-sinatra portfolio project" content="My sinatra powered Fwitter App">
+
+  <link rel="stylesheet" href="css/styles.css?v=1.0">
```

PR #720, opened 2018-04-09
Done

connorgallopo:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
 require './config/environment'
 
 class ApplicationController < Sinatra::Base
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+    @tweet = Tweet.find(params[:id])
+    if logged_in? && (@tweet.user == current_user)
+      @tweet.update(content: params['content'])
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    if logged_in? && params['content'] != ''
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+  post '/tweets/:id/edit' do
+    @tweet = Tweet.find(params[:id])
+    if logged_in? && (@tweet.user == current_user)
+      @tweet.update(content: params['content'])
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+  end
+
+  def current_user
+    User.find_by(id: session[:user_id])
   end
 
-end
\ No newline at end of file
+  def logged_in?
+    if current_user
```
weezwo: Avoid returning boolean literals, instead relying on the truthiness of a statement. For instance, this method could read:

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+      @user = User.create(params)
+      session[:user_id] = @user.id
+      redirect to '/tweets'
+    else
+      redirect to '/signup'
+    end
+  end
+
+  def current_user
+    User.find_by(id: session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+      @tweet.update(content: params['content'])
+    else
+      redirect to '/login'
+    end
+  end
+
+  post '/tweets/new' do
+    if logged_in? && params['content'] != ''
+      @user = current_user
+      @tweet = Tweet.create(params)
```
weezwo: Another way to do this is to make use of the `has_many` relationship on User:
```
```
Both approaches are good, but in some cases you may find this option useful. Typically, you'll want to create resources in one "step" when possible.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+    @user = User.find_by_slug(params[:slug])
+    erb :'/users/show'
+  end
+
+  get '/logout' do
+    session.clear if logged_in?
+    redirect to '/login'
+  end
+
+  post '/tweets/:id/edit' do
```
weezwo: Typically, the update route would be `post '/tweets/:id'`.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+      erb :'/tweets/show_tweet'
+    else
+      redirect to '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by(id: params[:id])
+      if @tweet.user == current_user
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,123 @@
...
+        erb :'/tweets/edit_tweet'
+      end
+    else
+      redirect to '/login'
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
+    @tweet.delete if logged_in? && (@tweet.user == current_user)
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #723, opened 2018-04-10
Done

seanoughton:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+    if @tweet.user == current_user
+      @tweet.delete
+      redirect '/tweets'
+    else
+      redirect '/tweets'
+    end
+  end
+
+  post '/tweets/new' do
+    if params[:tweet][:content].empty? || params[:tweet][:content].split.size < 1
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+    def logged_in? #returns true if logged in
+      !!session[:user_id]
+    end
+
+    def current_user
+      #tests to see if there is a session param called :user_id
+      #if there is one, that means that a user was created and saved to the database
+      #if so, this uses the find method on User to find that user instance
+      if session[:user_id]
+        @current_user = User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+      session[:user_id] = @user.id
+      redirect to '/tweets'
+    else
+      redirect to '/login'
+    end
+  end
+
+  helpers do
+    def logged_in? #returns true if logged in
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+    @tweet = Tweet.find(params[:id])
+    if @tweet.user == current_user
+      @tweet.delete
+      redirect '/tweets'
+    else
+      redirect '/tweets'
+    end
+  end
+
+  post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,151 @@ class ApplicationController < Sinatra::Base
...
+    if params[:tweet][:content].empty? || params[:tweet][:content].split.size < 1
+      redirect '/tweets/new'
+    else
+      @tweet = Tweet.create(params[:tweet])
```
weezwo: Another way to do this is to make use of the `has_many` relationship on User:
```
```

PR #729, opened 2018-04-15
Done

lekan20:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    if logged_in?
+      erb :'/tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    #CREATE: creates new instance of tweets (creates a new tweet)
+    if !params[:content].empty?
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      erb :'/tweets/show_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    #UPDATE: edit tweets contents
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:user_id] && @tweet.user_id == session[:user_id]
```
weezwo: Your `current_user` method could be useful here. That said, I don't see a lot of students checking the tweet's user id. Good job!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  patch '/tweets/:id/edit' do
+    #UPDATE: update tweet instance
+    if !params[:content].empty?
+      @tweet = Tweet.find_by_id(params[:id])
+      @tweet.content = (params[:content])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    else
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    #DELETE: delete and insatnce of tweet
+    #Chekcs that the creator of the tweet is deleting it
+    @tweet = Tweet.find_by_id(params[:id])
+    if session[:user_id] && @tweet.user_id == session[:user_id]
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #736, opened 2018-04-21
Done

kfecho30:


app/controllers/application_controller.rb
```diff
@@ -3,8 +3,137 @@
...
+    if logged_in?
+      erb :'tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content] != ""
+      @tweet = Tweet.create(params)
```
kkterai: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,137 @@
...
+      @tweet = Tweet.find(params[:id])
+      erb :'/tweets/show_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
```
kkterai: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?
Note:

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,137 @@
...
+      erb :'/tweets/edit_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  patch '/tweets/:id' do
+    tweet = Tweet.find(params[:id])
```
kkterai: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,137 @@
...
+    if params[:content] != ""
+      tweet.content = params[:content]
+      tweet.save
+    else
+      redirect "/tweets/#{tweet.id}/edit"
+    end
+  end
+
+  post '/tweets/:id/delete' do
+    @tweet = Tweet.find(params[:id])
```
kkterai: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #738, opened 2018-04-22
Done

Bechev:


app/controllers/application_controller.rb
```diff
@@ -3,8 +3,17 @@
 class ApplicationController < Sinatra::Base
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,17 @@
...
   end
 
-end
\ No newline at end of file
+  helpers do
+
+    def logged_in?
+       !!session[:username]
+    end
+
```
kkterai: A `current_user` method could be useful here.

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      redirect '/login'
+    end
+
+    get '/users/:username' do
+      @user = User.find_by(params[:user])
+      erb :"/tweets/show"
+    end
+
+  post '/signup' do
+    if params[:username]=="" || params[:email]=="" || params[:password]==""
```
kkterai: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,17 @@
...
+    set :session_secret, 'secret'
     set :public_folder, 'public'
     set :views, 'app/views'
   end
 
-end
\ No newline at end of file
+  helpers do
+
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+    else
+      redirect :'/login'
+    end
+  end
+
+  post '/tweets' do
+    if params[:content]==""
+      redirect "/tweets/new"
+    else
+      @tweet = Tweet.create(content: params[:content])
```
kkterai: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+      erb :"/tweets/show"
+    else
+      redirect :'/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
+      if @tweet && @tweet.user.username == session[:username]
```
kkterai: Good job checking that the current user is the tweet owner!

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+        else
+          redirect to '/tweets'
+        end
+      end
+    else
+      redirect to '/login'
+    end
+  end
+
+  delete '/tweets/:id/delete' do
```
kkterai: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #743, opened 2018-04-28
Done

missnadia:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,68 @@
+class TweetsController < ApplicationController
```
kkterai: Great job splitting your actions among multiple controllers!

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
 
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
+  helpers do
+    def current_user
+      @current_user = User.find_by_id(session[:user_id])
```

```ruby
def current_user
end
```

app/helpers/helpers.rb
```diff
@@ -0,0 +1,12 @@
+require 'sinatra/base'
+
```
kkterai: This code looks like it was intended to serve the purpose of the `current_user` and `logged_in?` methods that are already implemented in your ApplicationController. I would recommend refactoring to remove any unused code.

PR #744, opened 2018-04-28
Done

KimFoukaris:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,142 @@ class ApplicationController < Sinatra::Base
   configure do
```
kkterai: This controller is overloaded. Consider separating your concerns by splitting some of these actions into separate controllers (i.e. a users controller and a tweets controller).

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,142 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/signup' do
+    if logged_in?
+      redirect '/tweets'
+    end
+      erb :'/users/create_user'
+  end
+
+  post "/signup" do
+    if params[:username] == "" || params[:password] == "" || params[:email] == ""
```
kkterai: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,142 @@ class ApplicationController < Sinatra::Base
...
+        if @tweet.user_id == current_user.id
+          @tweet.destroy
+        end
+      end
+    end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,142 @@ class ApplicationController < Sinatra::Base
...
+
+    def current_user
+      User.find(session[:user_id])
+    end
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,142 @@ class ApplicationController < Sinatra::Base
...
+      else
+        redirect to "/login"
+      end
+    end
+
+    post '/tweets' do
+      @user = current_user
+        if params[:content] != ""
+          @tweet = Tweet.create(content: params[:content], user_id: @user.id)
+          @tweet.save
```
kkterai: You should not need `@tweet.save` since you used `Tweet.create...` above.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,142 @@ class ApplicationController < Sinatra::Base
...
+          redirect :'/tweets/new'
+        end
+      end
+
+    get '/tweets/:id/edit' do
+      if logged_in?
+        @user = current_user
+        @tweet=Tweet.find(params[:id])
+        erb :"/tweets/edit_tweet"
+      else
```
kkterai: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,142 @@ class ApplicationController < Sinatra::Base
...
+        erb :'/tweets/show_tweet'
+      else
+        redirect to "/login"
+      end
+    end
+
+    patch '/tweets/:id' do
+      @tweet = Tweet.find(params[:id])
+      @user=current_user
+      if logged_in?
```
kkterai: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form. Although you set `@user` to `current_user`, this does not provide a check that the edited tweets belong to `current_user`.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,142 @@ class ApplicationController < Sinatra::Base
...
+        end
+      else
+        redirect to '/login'
+      end
+    end
+
+    post '/tweets/:id/delete' do
+      if logged_in?
+        @tweet=Tweet.find(params[:id])
+        if @tweet.user_id == current_user.id
```
kkterai: Nice job checking that the tweet user id matches the current_user id!

PR #753, opened 2018-05-07
Done

oconnojb:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,123 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "password_security"
   end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,123 @@ class ApplicationController < Sinatra::Base
...
+    end
+
+  end
+
+  get '/tweets' do
+    logged_in? ? (erb :'tweets/tweets') : (redirect '/login?failed=yes')
+  end
+
+  post '/tweets' do
+    if !params[:content].empty?
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,123 @@ class ApplicationController < Sinatra::Base
...
+    erb :'users/show'
+  end
+
+  helpers do
+		def logged_in?
+			!!session[:user_id]
+		end
+
+		def current_user
+			User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,123 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  get '/users/:slug' do
+    @user = User.find_by_slug(params[:slug])
+    erb :'users/show'
+  end
+
+  helpers do
+		def logged_in?
+			!!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,123 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/tweets/:id' do
+    @tweet = Tweet.find(params[:id]) if logged_in?
+    logged_in? ? (erb :'tweets/show_tweet') : (redirect '/login?failed=yes')
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find(params[:id])
+      @tweet.user == current_user ? (erb :'tweets/edit_tweet') : (redirect "/users/#{current_user.slug?cant_edit=yes}")
```
cernanb: I don't see a lot of students checking the tweet's user. Good job!

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,123 @@ class ApplicationController < Sinatra::Base
...
+    else
+      redirect '/login?failed=yes'
+    end
+  end
+
+  post '/tweets/:id/edit' do
```
cernanb: convention for this route is a patch request to '/tweets/:id'
```
patch '/tweets/:id'
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,123 @@ class ApplicationController < Sinatra::Base
...
+    if !params[:content].empty?
+      @tweet = Tweet.find(params[:id])
+      @tweet.content = params[:content]
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,123 @@ class ApplicationController < Sinatra::Base
...
+      @tweet.save
+      redirect "/tweets/#{@tweet.id}"
+    else
+      redirect "/tweets/#{params[:id]/edit}?content_empty=yes"
+    end
+  end
+
+  post '/tweets/:id/delete' do
```
cernanb: Convention for this route is for this to be a delete request to '/tweets/:id'
```
delete '/tweets/:id'
```

cernanb (commented):
Hi John,

Regards,

Cernan

PR #757, opened 2018-05-09
Done

mstmari:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,15 @@
+require './config/environment'
+
+class TweetsController < ApplicationController
+
```
cernanb: Looks like the rest of the routes related to tweets can be moved from the application_controller to this tweets_controller.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,169 @@
...
+
+
+
+
+
+
+post '/tweets/new' do
+  is_logged_in?
+
+  if params[:content].empty?
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,169 @@
...
+
+end
+end
+
+
+
+
+
+
+post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,169 @@
...
+#<----------helpers-------->
+
+
+helpers do
+
+
+
+  def current_user
+    if session[:user_id]
+    @current_user ||= User.find(session[:user_id])
```
cernanb: Excellent use of memoization!

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,169 @@
...
+
+
+
+
+post '/tweets/new' do
+  is_logged_in?
+
+  if params[:content].empty?
+    redirect '/tweets/new'
+  else @tweet = Tweet.create(params)
```

``` ruby
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,169 @@
...
+    @tweet = Tweet.find_by_id(params[:id])
+    erb :'/tweets/show_tweet'
+  end
+end
+
+get '/tweets/:id/edit' do
+  if  !is_logged_in?
+    redirect '/login'
+  else
+    @tweet = Tweet.find_by_id(params[:id])
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,169 @@
...
+if
+  !is_logged_in?
+    redirect '/login'
+
+else
+  @tweet = Tweet.find(params[:id])
+    if @tweet.content.empty?
+    redirect "/tweets/#{@tweet.id}/edit"
+  end
+  @tweet.update(content: params[:content])
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,169 @@
...
+  redirect '/'
+end
+end
+
+delete '/tweets/:id/delete' do
+if
+    !is_logged_in?
+    redirect '/login'
+else
+  @tweet = Tweet.find_by_id(params[:id])
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #791, opened 2018-06-06
Done

goatcheese82:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,168 @@
-require './config/environment'
+require "./config/environment"
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,168 @@
...
+  get "/signup" do
+    if !logged_in?
+      erb :'users/create_user', locals: {message: "Please sign up before you sign in"}
+    else
+      redirect to "/tweets"
+    end
+  end
+
+  post "/signup" do
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
```
weezwo: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.I think

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,168 @@
...
+      erb :'tweets/show_tweet'
+    else
+      redirect to "/login"
+    end
+  end
+
+  get "/tweets/:id/edit" do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet && @tweet.user == current_user
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,168 @@
...
+      end
+    else
+      redirect to "/login"
+    end
+  end
+
+  delete "/tweets/:id/delete" do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet && @tweet.user == current_user
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #795, opened 2018-06-10
Done

jpkim921:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,27 @@
...
 
-end
\ No newline at end of file
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,27 @@
...
+    # need to have the two lines below to have access to the sessions hash
+    enable :sessions
+    set :session_secret, "secret"
   end
 
-end
\ No newline at end of file
+
+  helpers do
+    def logged_in?
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,27 @@
 require './config/environment'
 
 class ApplicationController < Sinatra::Base
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+  end
+
+  post '/tweets' do
+    if !logged_in?
+      redirect '/login'
+    elsif params[:content].empty?
+      redirect '/tweets/new'
+    else
+      @tweet = current_user.tweets.create(content: params[:content])
```
kkterai: Great job making use of the `has_many` relationship on User!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if !logged_in?
+      redirect '/login'
+    else
+      @tweet = Tweet.find(params[:id])
+      if @tweet.user_id == current_user.id
```
kkterai: I don't see a lot of students checking the tweet's user id. Good job!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+        erb :"/tweets/edit_tweet"
+      else
+        redirect "/tweets"
+      end
+    end
+  end
+
+  patch '/tweets/:id' do
+    @tweet = Tweet.find(params[:id])
```
kkterai: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

PR #796, opened 2018-06-10
Done

VictoriaMeng:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
   end
 
-end
\ No newline at end of file
+  get '/' do
+    erb :index
+  end
+
+  get '/signup' do
+    if session.has_key?(:id)
```
kkterai: You can use your `logged_in?` helper method here, which is preferable to referring to the session object directly.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+      @user = User.find(session[:id])
+      redirect "/tweets"
+    else
+      erb :signup
+    end
+  end
+
+  get '/tweets' do
+    if session.has_key?(:id)
```
kkterai: You can use your `logged_in?` helper method here, which is preferable to referring to the session object directly.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+      @user = User.find(session[:id])
+      erb :"tweets/tweets"
+    else
+      redirect "/login"
+    end
+  end
+
+  get "/login" do
+    if session.has_key?(:id)
```
kkterai: You can use your `logged_in?` helper method here, which is preferable to referring to the session object directly.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
...
+    redirect "/login"
+  end
+
+  get "/users/:slug" do
+    @user = User.find_by_slug(params[:slug])
+    erb :"/users/show"
+  end
+
+  get "/tweets/new" do
+    if session.has_key?(:id)
```
kkterai: You can use your `logged_in?` helper method here, which is preferable to referring to the session object directly.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,130 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "secret"
   end
```
kkterai: This controller is overloaded. Consider separating your concerns by splitting some of these actions into separate controllers (i.e. a users controller and a tweets controller).
Converse:

PR #798, opened 2018-06-12
Done

EPIC448:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,28 @@
 require './config/environment'
 
 class ApplicationController < Sinatra::Base
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    if !logged_in?
+      erb :'users/create_user', locals: {message: "Please sign up before you sign in"}
+    else
+      redirect to '/tweets'
+    end
+  end
+
+  post '/signup' do
+
+    if params[:username] == "" || params[:email] == "" || params[:password] == ""
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+  end
+
+  post '/tweets' do
+    @tweet = Tweet.new(params)
+    @user = current_user
+    #check if tweet is empty.
+    # This is where we set the name for song/ it want us to pass in an hash.
+    if logged_in? && !@tweet.content.blank? && @tweet.save
+      @user.tweets << @tweet
```
weezwo: The tweet instantiation and user association can both be accomplished in the same step:

``` ruby
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+      redirect "/login"
+    end
+  end
+
+  post '/tweets' do
+    @tweet = Tweet.new(params)
+    @user = current_user
+    #check if tweet is empty.
+    # This is where we set the name for song/ it want us to pass in an hash.
+    if logged_in? && !@tweet.content.blank? && @tweet.save
```
weezwo: Great job using ActiveRecord model validations! If you validate the presence of content in a tweet, you don't have to include a controller validation like `!@tweet.content.blank?`

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+  end
+
+
+  get '/tweets/:id/edit' do
+
+     if logged_in?
+    @tweet = Tweet.find_by_id(params[:id]) # slug helps to find by name instaed of ID
+    erb :'tweets/edit_tweet'
```
weezwo: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+  end
+
+  # <this connect here
+
+  patch "/tweets/:id" do
+    @tweet = Tweet.find(params[:id])
+
+        if logged_in? && !params[:content].blank?
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+        redirect to "/tweets/#{@tweet.id}/edit"
+      end
+
+    end
+
+
+  delete "/tweets/:id/delete" do
+    @tweet = Tweet.find_by_id(params[:id])
+
+    if logged_in? && @tweet.user == current_user
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #799, opened 2018-06-12
Project submitted

wlycdgrfromflatiron:


app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,130 @@
+class TweetController < ApplicationController
```

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        session.delete(:login_error) ?
+            @login_error_message = session.delete(:login_error_message) :
+            @login_error_message = ''
+
+        erb :'/users/login'
+    end
+
+    post '/login' do
+        if params[:username].empty?
```
weezwo: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,19 @@ class ApplicationController < Sinatra::Base
...
+  get '/' do
+    erb :index
+  end
+
+  def logged_in?
+    !!session[:user_id]
+  end
+
+  def user
+    User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,19 @@ class ApplicationController < Sinatra::Base
...
+    enable :sessions
+    set :session_secret, "not that secret but, you know, this isn't the point of this lab"
   end
 
+  get '/' do
+    erb :index
+  end
+
+  def logged_in?
+    !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+    end
+
+    post '/tweets' do
+        if !logged_in?
+            redirect to '/login'
+        end
+
+        if params[:content].empty?
+            session[:new_tweet_error] = true
+            session[:new_tweet_error_message] = 'Tweets must not be blank!'
```
weezwo: Prefer to put errors messages into flash messages rather than the session!

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+            redirect to '/tweets/new'
+        end
+
+        tweet = Tweet.create(content: params[:content])
+        tweet.user = user
```
weezwo: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```

app/controllers/tweet_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+            redirect to '/login'
+        end
+
+        @tweet = Tweet.find(params[:id])
+
+        if !@tweet
+            redirect to '/tweets'
+        end
+
+        if !user.tweets.include?(@tweet)
```
weezwo: I don't see a lot of students checking the tweet's user id. Good job!

PR #802, opened 2018-06-14
Done

kanndide:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  get '/logout' do
+  	session.clear
+    redirect "/login"
+  end
+
+  get '/tweets' do
+  	if logged_in?
+  		@user = User.find_by_id(session[:user_id])
```
weezwo: Your `current_user` method could be useful here.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
   configure do
     set :public_folder, 'public'
     set :views, 'app/views'
+    enable :sessions
+    set :session_secret, "password_security"
+  end
+
+  get '/' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
...
+  		redirect '/tweets'
+  	else
+		erb :'users/create_user'
+	end
+  end
+
+  post "/signup" do
+    @user = User.create(params)
+
+    if @user.save && params[:username] != "" && params[:password] != ""
```
weezwo: Great job using ActiveRecord model validations! They will take the place of manual checks in the controller.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
...
+  	end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
+    end
+
+    def current_user
+      User.find(session[:user_id])
```

```ruby
def current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
...
+  		@tweet.delete
+  		redirect "/users/#{@user.slug}"
+  	else
+  		redirect '/login'
+  	end
+  end
+
+  helpers do
+    def logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
...
+
+  get '/tweets/new' do
+  	if logged_in?
+  		erb :'tweets/create_tweets'
+  	else
+  		redirect '/login'
+  	end
+  end
+
+  post '/tweets/new' do
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
...
+  	if params[:content] != ""
+	  	@user = User.find_by_id(session[:user_id])
+	  	@tweet = Tweet.create(params)
+	  	@tweet.user = @user
```
weezwo: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
...
+  		erb :'/tweets/show_tweet'
+  	else
+  		redirect '/login'
+  	end
+  end
+
+  get '/tweets/:id/edit' do
+  	if logged_in?
+  		@tweet = Tweet.find_by_id(params[:id])
+  		erb :'/tweets/edit_tweet'
```
weezwo: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
...
+  	else
+  		redirect '/login'
+  	end
+  end
+
+  patch '/tweets/:id' do
+  	@tweet = Tweet.find_by_id(params[:id])
+  	if params[:content] != ""
+	  	@tweet.content = params[:content]
```
weezwo: I would contend that you still want to double check the user's association to the tweet here. There are other ways to submit a PATCH/POST request than going through an HTML form...

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,149 @@ class ApplicationController < Sinatra::Base
...
+	  	redirect "/tweets/#{@tweet.id}"
+	else
+		redirect "/tweets/#{@tweet.id}/edit"
+	end
+  end
+
+  delete '/tweets/:id/delete' do
+  	@user = User.find_by_id(session[:user_id])
+  	@tweet = Tweet.find_by_id(params[:id])
+  	if logged_in? && @user.id == @tweet.user_id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

kanndide: This one took me a second but I figured it out in the end. 
I just had to change the redirect path on the example you put from "/tweets/#{tweet.id}" to "/tweets/#{params[:id]}". If the tweet variable is nil it wouldn't redirect. Was that an acceptable way to write out the path? I appreciate the code review and have attempted to go over them all to fix/understand them.

PR #807, opened 2018-06-18
Done

emikojima:


app/controllers/application_controller.rb
```diff
@@ -3,8 +3,148 @@
 class ApplicationController < Sinatra::Base
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,148 @@
...
+  end
     set :public_folder, 'public'
     set :views, 'app/views'
+
+    get '/' do
+      erb :index
+    end
+
+    def logged_in?
+      !!session[:user_id]
```

```ruby
def logged_in?
  !!current_user
end
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,148 @@
...
+    get '/signup' do
+      if self.logged_in?
+          redirect "/tweets"
+      else
+        erb :'/users/create_user'
+      end
+    end
+
+    post '/signup' do
+      if params[:username]=="" || params[:email]==""|| params[:password]==""
```

app/controllers/application_controller.rb
weezwo: That said, you will also want to define a `current_user` method!

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,148 @@
...
+
+    get '/tweets/new' do
+      if self.logged_in?
+        erb :'/tweets/create_tweet'
+      else
+        redirect "/login"
+      end
+      end
+
+    post '/tweets/new' do
```
weezwo: Typically, following [RESTful convention](https://learn.co/tracks/full-stack-web-development-v5/sinatra/activerecord/sinatra-restful-routes), tweet creation would be posted to `/tweets` instead of `/tweets/new`.

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,148 @@
...
+      @user = User.find(session[:user_id])
+      if self.logged_in? && params[:content] != ""
+        @user.tweets << Tweet.new(content: params[:content])
```
weezwo: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,148 @@
...
+        redirect "/users/#{@user.slug}"
+      else
+        redirect "/tweets/new"
+      end
+    end
+
+    get '/tweets/:id/edit' do
+      if logged_in?
+        @tweet = Tweet.all.find(params[:id])
+        if @tweet && session[:user_id] == @tweet.user_id
```

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,148 @@
...
+        erb :'tweets/show_tweet'
+      else
+        redirect "/login"
+      end
+    end
+
+
+  post '/tweets/:id/delete' do
+    @tweet = Tweet.all.find(params[:id])
+    if logged_in? && session[:user_id] == @tweet.user_id
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #810, opened 2018-06-19
Done

eyi1:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,22 @@ class ApplicationController < Sinatra::Base
   configure do
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    get '/signup' do 
+        if !logged_in?             
+            erb :'/users/create_user'
+        else
+            redirect to '/tweets'
+        end
+    end
+
+    post '/signup' do
+        if params[:username] == "" || params[:email] == "" || params[:password] == ""
```

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,22 @@ class ApplicationController < Sinatra::Base
...
+    erb :index
+  end
+
+  helpers do
+    def logged_in?
+      !!current_user
+    end
+
+    def current_user
+      User.find_by(session[:user_id]) if session[:user_id]
```

```ruby
def current_user
end
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+        else
+            redirect to '/login'
+        end
+    end
+
+    post '/tweets' do
+        if logged_in?
+            if !params[:content].empty? 
+                @tweet = Tweet.create(content: params[:content])
+                current_user.tweets << @tweet #current_user.tweet.build(content: params[:content])
```
weezwo: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```

weezwo: Your commented out attempt actually looks great! Looks like all you're missing is an 's': ```
```
Because a User has_many Tweets, ActiveRecord will give any user a method `tweets` to view those tweets belonging to the user. It is this method that `#build` is built from.

File: app/controllers/tweets_controller.rb [hunk 5]
```diff
@@ -0,0 +1,85 @@
...
+            end
+         else 
+            redirect to '/login'
+         end
+    end
+
+    get '/tweets/:id/edit' do
+        if logged_in?
+          @tweet = Tweet.find_by_id(params[:id])
+          if @tweet && @tweet.user == current_user
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+        else
+          redirect to '/login'
+        end
+      end
+    
+    patch '/tweets/:id' do 
+        if logged_in?
+            @tweet = Tweet.find_by_id(params[:id])
+            if !params[:content].empty? 
+                @tweet.update(content: params[:content])
```

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+            end
+        else
+            redirect to '/login'
+        end
+    end 
+
+    delete '/tweets/:id/delete' do 
+        if logged_in?
+         @tweet = Tweet.find_by_id(params[:id])
+          if @tweet && @tweet.user == current_user
```

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

PR #811, opened 2018-06-20
Done

njpowers7915:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,88 @@
 require './config/environment'
+require 'pry'
 
 class ApplicationController < Sinatra::Base
```
howardbdev: This controller is overloaded. Consider separating your concerns by splitting some of these actions into separate controllers (i.e. a users controller and a tweets controller, which you have but are not using).

Also, this is where your configurations are set, and all other controllers inherit from this one, so there is no need for the `register`, `set :views`, or `configure do #...` anywhere other than here.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+    register Sinatra::ActiveRecordExtension
+    set :views, Proc.new { File.join(root, "../views/") }
+
+    configure do
+      enable :sessions
+      set :session_secret, "secret"
+    end
+
+
+end
```
howardbdev: These configurations are set up in ApplicationController, so all we need here are the routes

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,84 @@
+class SessionsController < ApplicationController
+    register Sinatra::ActiveRecordExtension
+    set :views, Proc.new { File.join(root, "../views/") }
+
+    configure do
+      enable :sessions
+      set :session_secret, "secret"
+    end
```
howardbdev: We don't need to set up configurations like this outside of ApplicationController

app/helpers/helpers.rb
```diff
@@ -0,0 +1,10 @@
+class Helpers
+    def self.current_user(session)
+        @user = User.find_by(id: session[:user_id])
+    end
+
+    def self.is_logged_in?(session)
+        @user = User.find_by(id: session[:user_id])
+        !@user.nil?
+    end
+end
```
howardbdev: We could just drop this helper methods into the ApplicationController, in a `helpers do ... end` block, like so:

```
# in class ApplicationController
helpers do
  def self.current_user(session)
      @user = User.find_by(id: session[:user_id])
  end

  def self.is_logged_in?(session)
     @user = User.find_by(id: session[:user_id])
     !@user.nil?
  end
end
```

app/views/tweets/new.erb
```diff
@@ -0,0 +1,4 @@
+<form action="/tweets/new" method="POST">
```
howardbdev: RESTful convention for post routes should go to just the plural of the model, in this case, `POST '/tweets'`

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,88 @@
...
+  get '/tweets/new' do
+      @user = Helpers.current_user(session)
+      if Helpers.is_logged_in?(session) == false
+          redirect '/login'
+      else
+          erb :'tweets/new'
+      end
+  end
+
+  post '/tweets/new' do
```
howardbdev: Typically, following [RESTful convention](https://learn.co/tracks/full-stack-web-development-v5/sinatra/activerecord/sinatra-restful-routes), tweet creation would be posted to `/tweets` instead of `/tweets/new`.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,88 @@
...
+      if params[:content] == ""
```
howardbdev: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.
This applies anywhere you're creating or updating and checking that the input isn't empty.

app/models/tweet.rb
```diff
@@ -0,0 +1,3 @@
+class Tweet < ActiveRecord::Base
+    belongs_to :user
```
howardbdev: An ActiveRecord validation for content would look like this:
```
validates :content, presence: true
```

app/models/user.rb
```diff
@@ -0,0 +1,12 @@
+class User < ActiveRecord::Base
+    has_many :tweets
+    has_secure_password
+
```
howardbdev: We could also use validations on the User model, such as:
```
validates :username, :email, presence: true
validates :username, :email, uniqueness: true
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,88 @@
...
+      end
+  end
+
+  delete '/tweets/:id/delete' do
+      @user = Helpers.current_user(session)
+      if Helpers.is_logged_in?(session) == false
+        redirect '/login'
+      else
+          @tweet = Tweet.find_by(id: params[:id])
+          if @tweet.user_id == session[:user_id]
```
howardbdev: Here, we could save lookup time by *only* checking the current user's tweets:
```
if @tweet
  #...
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,88 @@
...
+    end
+  end
+
+  patch '/tweets/:id' do
+      @tweet = Tweet.find_by(id: params[:id])
+      if params[:content] == ""
+          redirect "tweets/#{@tweet.id}/edit"
+      else
+          @tweet.content = params["content"]
+          @tweet.save
```
howardbdev: Just use `#update`:
```
```
In the case that we switch to AR validations, it could look something like:
```
if @tweet.update(content: params[:content])
  # success
else
  # redirect or render edit form
end
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,88 @@
...
+            erb :'tweets/edit'
+        else
+            erb :error
+        end
+    end
+  end
+
+  patch '/tweets/:id' do
+      @tweet = Tweet.find_by(id: params[:id])
+      if params[:content] == ""
```
howardbdev: Another good place to just look for the current user's tweets, since we should only be modifying tweets we own:
```
if @tweet
  #...
```

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,88 @@
 require './config/environment'
+require 'pry'
```
howardbdev: With pry in our Gemfile, we shouldn't need to require it here also.

PR #812, opened 2018-06-20
Done

alyssa0528:


app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+        redirect '/login'
+      end
+    end
+
+#process new tweet form submission based on params
+  post '/tweets' do
+    if logged_in? #if logged in...
+      if !params[:content].empty? #and if tweet_content isn't empty...
+        @tweet = current_user.tweets.build(params) #create new tweet instance via #build
+        @tweet.save #save new tweet instance to database
```
howardbdev: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

In this case, we could lose the `if !params[:content].empty? ` statement entirely, add the validation for content in the Tweet model, and use this pattern:
```
if @tweet.save # If @tweet is invalid it won't save
  redirect "/tweets/#{@tweet.id}"
else
  redirect '/tweets/new' 
end
```

This comment also applies to the patch and delete routes.

app/controllers/tweets_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+    else
+      redirect '/login' #if not logged in, go to login page
+    end
+  end
+
+#show edit tweet page for a specific tweet
+  get '/tweets/:id/edit' do
+    if logged_in?
+      @tweet = Tweet.find_by_id(params[:id])
+      if @tweet.user == current_user
```
howardbdev: We could eliminate some lookup time here by only searching the current user's tweets, and then use just @tweet to control the following `if`:
```
if @tweet
  # ...
```
This comment also applies to the patch and delete routes.

app/models/tweet.rb
```diff
@@ -0,0 +1,3 @@
+class Tweet < ActiveRecord::Base
+  belongs_to :user
+end
```
howardbdev: The previously mentioned validation would look like this:
```
validates :content, presence: true
```

app/models/user.rb
```diff
@@ -0,0 +1,14 @@
+class User < ActiveRecord::Base
+  has_many :tweets
+  has_secure_password
+
```
howardbdev: It might also be useful to add more validations to the User model, such as presence and uniqueness for username and email:
```
validates :username, :email, presence: true
validates :username, :email, uniqueness: true
```
Then, in controller actions that create or update these attributes, we can use `@user.valid?` or `if @user.save` to control the next steps.

PR #820, opened 2018-06-24
Done

Almighty-Mose:


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+  helpers do
+    def logged_in?
+      !!current_user
+    end
+
+    def current_user
+      @current_user ||= User.find_by(id: session[:user_id]) if session[:user_id]
+    end
+  end
 end
```

app/controllers/tweets_controller.rb
```diff
@@ -1,4 +1,73 @@
...
 
+  get '/tweets/new' do
+    if logged_in?
+      erb :'tweets/create_tweet'
+    else
+      redirect '/login'
+    end
+  end
 
+  post '/tweets/new' do
```
kkterai: This action should be `post '/tweets`. `get '/tweets/new'` does not typically have a corresponding `post '../new'` route.

app/controllers/tweets_controller.rb
```diff
@@ -1,4 +1,73 @@
...
+    if !params[:content].empty?
+      @tweet = Tweet.create(params)
+      @tweet.user_id = current_user.id
+      @tweet.save
+
```
kkterai: The tweet creation and user association can both be accomplished in the same step:

``` ruby
```
In addition, `@tweet.save` is not necessary as `@tweet` has already been persisted when it was `create`d.

app/controllers/tweets_controller.rb
```diff
@@ -1,4 +1,73 @@
...
+      @tweet = Tweet.find_by_id(params[:id])
+
+      erb :'tweets/show_tweet'
+    else
+      redirect '/login'
+    end
+  end
+
+  get '/tweets/:id/edit' do
+    if logged_in?
```
kkterai: Shouldn't you verify a tweet belongs to the `current_user` (and not just that any user is logged in) before allowing them to edit?

app/controllers/tweets_controller.rb
```diff
@@ -1,4 +1,73 @@
...
+    else
+      redirect '/login'
+    end
+  end
+
+  patch '/tweets/:id/edit' do
+    @tweet = Tweet.find_by_id(params[:id])
+    if !params[:content].empty?
+      @tweet.content = params[:content]
+      @tweet.save
```

app/controllers/tweets_controller.rb
```diff
@@ -1,4 +1,73 @@
...
+
+      redirect '/tweets'
+    else
+      redirect "/tweets/#{@tweet.id}/edit"
+    end
+  end
+
+  delete '/tweets/:id/delete' do
+    @tweet = Tweet.find_by_id(params[:id])
```
kkterai: Prefer to find tweets that are already only associated to the `current_user` than to manually check the `user_id` foreign key of the tweet against the `current_user.id`

```ruby
if tweet && tweet.destroy
  redirect "/tweets"
else
end
```

app/controllers/users_controller.rb
```diff
@@ -1,4 +1,52 @@
...
+    if @user != nil && @user.authenticate(params[:password])
+      session[:user_id] = @user.id
+      redirect '/tweets'
+    else
+      redirect '/signup'
+    end
+  end
+
+  post '/signup' do
+    if !params[:username].empty? && !params[:email].empty? && !params[:password].empty?
```
kkterai: I think you could use [ActiveRecord validations](http://guides.rubyonrails.org/active_record_validations.html) (e.g. `validates_presence_of :username`) to perform these types of checks more easily.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,21 @@ class ApplicationController < Sinatra::Base
...
+  end
+
+  helpers do
+    def logged_in?
+      !!current_user
+    end
+
+    def current_user
+      @current_user ||= User.find_by(id: session[:user_id]) if session[:user_id]
+    end
```
kkterai: Good job memoizing `current_user`!

PR #1011, opened 2018-12-06
Project submitted

musicmills1990:


musicmills1990:
I'm fairly happy with this, obviously it could use a lot of styling with css but it's user friendly. Things that could improve: 
-User's listed on the tweets page
-The flash[:errors] for trying to edit someone else's tweet doesn't work.

musicmills1990:
one more bug, you shouldn't be able to submit an empty tweet through your own edit page.

